PR #4, opened 2018-04-08
Add missing features used in gluon's source mapping

Marwes:
PR name is a bit blatant but I have been updating gluon to use codemap instead and found a few things that were missing (compared to the similar structure previously used in gluon). I think most things should be pretty straightforward improvements though naming may need to be changed.

https://github.com/gluon-lang/gluon/pull/505

### DIFF
--- a/codespan/src/filemap.rs
+++ b/codespan/src/filemap.rs
@@ -145,6 +145,22 @@ impl FileMap {
         self.span
     }
 
+    pub fn offset(&self, line: LineIndex, column: ColumnIndex) -> Option<ByteOffset> {
+        self.line_offset(line).ok().and_then(|mut offset| {
+            offset += ByteOffset::from(column.0 as i64);
+            if offset.to_usize() >= self.src.len() {
+                None
+            } else {
+                Some(offset)
+            }
+        })
+    }
+
+    pub fn byte_index(&self, line: LineIndex, column: ColumnIndex) -> Option<ByteIndex> {
+        self.offset(line, column)
+            .map(|offset| self.span.start() + offset)
+    }
+
     /// Returns the byte offset to the start of `line`
     pub fn line_offset(&self, index: LineIndex) -> Result<ByteOffset, LineIndexError> {
         self.lines
--- a/codespan/src/filemap.rs
+++ b/codespan/src/filemap.rs
@@ -73,6 +73,17 @@ pub enum ByteIndexError {
 }
 
 #[derive(Debug, Fail, PartialEq)]
+pub enum LocationError {
+    #[fail(display = "Line out of bounds - given: {:?}, max: {:?}", given, max)]
+    LineOutOfBounds { given: LineIndex, max: LineIndex },
+    #[fail(display = "Column out of bounds - given: {:?}, max: {:?}", given, max)]
+    ColumnOutOfBounds {
+        given: ColumnIndex,
+        max: ColumnIndex,
+    },
+}
+
+#[derive(Debug, Fail, PartialEq)]
 pub enum SpanError {
     #[fail(display = "Span out of bounds - given: {}, span: {}", given, span)]
     OutOfBounds { given: ByteSpan, span: ByteSpan },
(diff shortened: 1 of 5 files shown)


codespan/src/filemap.rs
```diff
@@ -117,6 +145,22 @@ impl FileMap {
         self.span
     }
 
+    pub fn offset(&self, line: LineIndex, column: ColumnIndex) -> Option<ByteOffset> {
+        self.line_offset(line).ok().and_then(|mut offset| {
+            offset += ByteOffset::from(column.0 as i64);
```
brendanzab: Should we have an error if this runs into the next line?

brendanzab: Kind of weird it is not called `byte_offset`, but then it would be inconsistent with `line_offset` 🤔

codespan/src/filemap.rs
```diff
@@ -15,6 +15,30 @@ pub enum FileName {
...
+    fn from(name: String) -> FileName {
+        FileName::virtual_(name)
+    }
+}
+
+impl From<&'static str> for FileName {
+    fn from(name: &'static str) -> FileName {
+        FileName::virtual_(name)
+    }
+}
```
brendanzab: Awesome stuff. I wasn't sure about adding these, but I think it's fine.

codespan/src/filemap.rs
```diff
@@ -68,6 +92,10 @@ pub struct FileMap {
 }
 
 impl FileMap {
+    pub fn anonymous(src: String) -> FileMap {
+        Self::new(FileName::virtual_(""), src, ByteIndex::from(1))
```
brendanzab: When is this used?

codespan/src/filemap.rs
Marwes: > Should we have an error if this runs into the next line?

Probably should have an error for that and to capture the error from `line_offset` as well. It gets a bit verbose when having precise errors.

codespan/src/filemap.rs
Marwes: Yeah, I don't think they should be that surprising and it was a bit tedious to need to import `FileName` frequently when I really just want to give a name to files added with https://github.com/brendanzab/codespan/blob/ccff649894f5ba3ce8c1c4c823a9ca2798d90ec8/codespan/src/codemap.rs#L31

codespan/src/filemap.rs
Marwes: Just for tests in my case. Maybe we want to make it `fn new(name: FileName, src: String) -> FileMap` though. I think there is merit to be able to construct standalone `FileMap`s in any case.

brendanzab: Hum - yeah, I think what I was trying to avoid was invalid spans being created, like codemap does. But maybe I could just give up on that. 🤔

Marwes: It is still easy to create two codemaps and accidentally use spans from one in the other though. Not really a solvable problem unfortunately :(

codespan/src/span.rs
```diff
@@ -9,7 +10,7 @@ pub struct Span<I> {
     end: I,
 }
 
-impl<I: Index> Span<I> {
+impl<I: Ord> Span<I> {
```
brendanzab: Why this change?

brendanzab: I guess I was trying to limit it to fixed locations, rather than potentially folks using it for offsets. But I can see you reasoning, so let's go with that.

Marwes: Okay, I figured it was just out of convenience that all functions were bounded by `I: Index`.  I could implement `Index` for my `Location` type and restore this bound so that guarantee remains instead (though `start` and `end` ought to still stay unbounded)

brendanzab: Yeah, have a look - it might be a pain to implement for `Location` - eg. you might need to then introduce a corresponding `LocationOffset` type. I understand if that's a bit much to ask for though!

Marwes:
This should be good now from my side at least

codespan/src/filemap.rs
brendanzab: Could we perhaps remove this and just make the `new` constructor public?

Marwes: Done. I am a bit worried about exposing `start: ByteIndex` as it is not obvious that `0` is is not a valid index.  I hid that part of the public API so that it is only possible to create standalone `FileMap`s without `CodeMap`.

Issue #1, opened 2018-02-20
Improve error formatting for CLI applications


brendanzab:
Currently I'm not doing much about this. Here are some options

- We could try to follow Rust closely.
- Get inspiration from other languages and go our own way. 
- Provide multiple error formatting styles (could help us support other backends, like the [Language Server Protocol](https://microsoft.github.io/language-server-protocol/)).

I'd also like to be able to support domain-specific errors, such as those produced by LALRPOP - at least providing an escape hatch to allow them to be easily created!

# Inspiration:

Here are some examples of what the output might look like:

## [Reason](https://reasonml.github.io/community/blog/#way-way-waaaay-nicer-error-messages)

[Source code](https://github.com/BuckleScript/bucklescript/tree/master/jscomp/super_errors)

![Reason error output](https://user-images.githubusercontent.com/1909539/29709301-ab04eac6-8940-11e7-8d2b-c65f808b6be8.png)

![Reason warning output](https://user-images.githubusercontent.com/1909539/29711739-431be094-894b-11e7-87a6-bc1d6aeea043.png)

![Reason React error](https://user-images.githubusercontent.com/1909539/29712284-f1013bb2-894d-11e7-9596-1cca54d5c331.png)

## [LALRPOP](http://smallcultfollowing.com/babysteps/blog/2016/03/02/nice-errors-in-lalrpop/)

[Source code](https://github.com/nikomatsakis/lalrpop/tree/master/lalrpop/src/message)

```
calc.lalrpop:6:5: 6:34: Ambiguous grammar detected

  The following symbols can be reduced in two ways:
    Expr "*" Expr "*" Expr

  They could be reduced like so:
    Expr "*" Expr "*" Expr
    ├─Expr──────┘        │
    └─Expr───────────────┘

  Alternatively, they could be reduced like so:
    Expr "*" Expr "*" Expr
    │        └─Expr──────┤
    └─Expr───────────────┘

  Hint: This looks like a precedence error related to `Expr`. See the LALRPOP
  manual for advice on encoding precedence.
```

## [Elm](http://elm-lang.org/blog/compilers-as-assistants)

[Source code](https://github.com/elm-lang/elm-compiler/tree/master/src/Reporting)

![Elm big record error](http://elm-lang.org/assets/blog/error-messages/0.16/big-record.png)

![Elm error](http://elm-lang.org/assets/blog/error-messages/0.16/context.png)

[...]

PR #5, opened 2018-04-11
Add colorized output for emit

Marwes:
codespan

I haven't spent to much time on this but but I think this should serve as a starting point for #1 (or at least a start/continuation of discussion on it). I would like to get something like this in in any case so that I don't lose out on the source display part in https://github.com/gluon-lang/gluon/pull/505 :)

### DIFF
--- a/codespan-reporting/src/emitter.rs
+++ b/codespan-reporting/src/emitter.rs
@@ -2,7 +2,7 @@ use std::io;
 
 use codespan::CodeMap;
 
-use termcolor::{ColorSpec, WriteColor};
+use termcolor::{Color, ColorSpec, WriteColor};
 
 use Diagnostic;
 
@@ -21,7 +21,6 @@ where
             },
             Some(file) => {
                 let (line, col) = file.location(label.span.start()).expect("location");
-
                 writeln!(
                     writer,
                     "- {}:{}:{}",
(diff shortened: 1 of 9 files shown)


Marwes:
Diff https://github.com/brendanzab/codespan/pull/5/files/56d4840a6f9ad770bffb54f56b8ce091c5b6cd1b..fffbb382086c3a97a8030130939fdc5a1d2965bf

codespan/src/span.rs
```diff
@@ -84,11 +84,11 @@ impl<I: Index> Span<I> {
...
     /// let span = Span::new(ByteIndex(3), ByteIndex(6));
-    /// assert_eq!(span.with_lo(ByteIndex(2)), Span::new(ByteIndex(2), ByteIndex(6)));
-    /// assert_eq!(span.with_lo(ByteIndex(5)), Span::new(ByteIndex(5), ByteIndex(6)));
-    /// assert_eq!(span.with_lo(ByteIndex(7)), Span::new(ByteIndex(6), ByteIndex(7)));
+    /// assert_eq!(span.with_start(ByteIndex(2)), Span::new(ByteIndex(2), ByteIndex(6)));
+    /// assert_eq!(span.with_start(ByteIndex(5)), Span::new(ByteIndex(5), ByteIndex(6)));
+    /// assert_eq!(span.with_start(ByteIndex(7)), Span::new(ByteIndex(6), ByteIndex(7)));
     /// ```
-    pub fn with_lo(self, start: I) -> Span<I> {
+    pub fn with_start(self, start: I) -> Span<I> {
```
brendanzab: 👏

codespan-reporting/src/emitter.rs
```diff
@@ -1,23 +1,92 @@
...
+
 use codespan::CodeMap;
 
+use termcolor::{Color, ColorSpec, WriteColor};
+
 use Diagnostic;
 
-pub fn emit(codemap: &CodeMap, diagnostic: &Diagnostic) {
-    println!("{}: {}", diagnostic.severity, diagnostic.message);
+pub fn emit<W>(mut writer: W, codemap: &CodeMap, diagnostic: &Diagnostic) -> io::Result<()>
```
brendanzab: Goood. Was wondering if I should switch it to be `eprintln!`, but this is better!

brendanzab:
This is awesome! Super looking forward to it! 😍

codespan-reporting/src/emitter.rs
brendanzab: So is `WriteColor` configurable? ie. my repl has a `--color=[auto|always|never]` flag - would I be able to set that?

Marwes: Yup, you just set `ColorChoice` when you create the `WriteColor` instance https://github.com/brendanzab/codespan/pull/5/files#diff-a45cab6253488a1ff30335b30b8dc71aR38

brendanzab: niiiice

PR #6, opened 2018-04-12
Hide the ByteIndex parameter for FileMap in the public API

Marwes:
It is not clear that 0 is not a valid index and it is probably not useful for external users to create FileMap with non-1 indexes without CodeMap.

### DIFF
--- a/codespan/src/filemap.rs
+++ b/codespan/src/filemap.rs
@@ -103,8 +103,13 @@ pub struct FileMap {
 }
 
 impl FileMap {
-    /// Construct a new filemap, creating an index of line start locations
-    pub fn new(name: FileName, src: String, start: ByteIndex) -> FileMap {
+    /// Construct a new, standalone filemap. Can be usefol for tests which consists of a single
+    /// source file
+    pub fn new(name: FileName, src: String) -> FileMap {
+        FileMap::with_index(name, src, ByteIndex(1))
+    }
+
+    pub(crate) fn with_index(name: FileName, src: String, start: ByteIndex) -> FileMap {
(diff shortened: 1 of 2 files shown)


codespan/src/filemap.rs
```diff
@@ -104,7 +104,11 @@ pub struct FileMap {
 
 impl FileMap {
     /// Construct a new filemap, creating an index of line start locations
```
brendanzab: I wonder if we could update the documentation here, saying that this is hand for tests, but production code should probably go via a `CoadeMap` 🤔

brendanzab:
Yup, good call. Just think we could improve the docs here. Not sure though.

Marwes:
Done

codespan/src/filemap.rs
```diff
@@ -103,8 +103,13 @@ pub struct FileMap {
 }
 
 impl FileMap {
-    /// Construct a new filemap, creating an index of line start locations
-    pub fn new(name: FileName, src: String, start: ByteIndex) -> FileMap {
+    /// Construct a new, standalone filemap. Can be usefol for tests which consists of a single
+    /// source file
```
brendanzab: Spelling mistake! 😅  How about:

```rust
/// Construct a new, standalone filemap.
///
/// This can be useful for tests that consist of a single source file. Production code should however 
/// use `CodeMap::add_filemap` or `CodeMap::add_filemap_from_disk` instead.
```

Marwes: Ooops, that is better still though.

brendanzab:
Thanks!

PR #11, opened 2018-04-14
Add some missing impls

Marwes:
For the `Clone` bound, I want to be able to store a `CodeMap` inside some of my error types so that they can implement `fmt::Display` and still print out the source positions.

### DIFF
--- a/codespan/src/codemap.rs
+++ b/codespan/src/codemap.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use filemap::{FileMap, FileName};
 use index::{ByteIndex, ByteOffset};
 
-#[derive(Debug, Default)]
+#[derive(Clone, Debug, Default)]
 pub struct CodeMap {
     files: Vec<Arc<FileMap>>,
 }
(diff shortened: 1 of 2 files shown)


codespan/src/codemap.rs
```diff
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use filemap::{FileMap, FileName};
 use index::{ByteIndex, ByteOffset};
 
-#[derive(Debug, Default)]
+#[derive(Clone, Debug, Default)]
 pub struct CodeMap {
     files: Vec<Arc<FileMap>>,
```
brendanzab: I wonder if the `Vec` could also be `Arc`ed? Might have to revisit what Rust does - I think I simplified things here.

brendanzab:
Good idea!

codespan/src/codemap.rs
Marwes: It could and then one would use `make_mut` to get a unique reference to mutate it. But that could be done as `Arc<CodeMap>` as well so.

PR #14, opened 2018-04-16
Fix marker padding

brendanzab:
I broke the padding in c1a382ef6e6ddb9bbd8d0f338e5887fdb7b7d2ed due to my misunderstanding of what was going on. I've created a new `Pad` datatype to make it clearer. 

Following rustc's lead I've also:

- made the secondary labels use a different styling to the primary labels, 
- used blue coloring rather than cyan for the line numbers

### DIFF
--- a/codespan-reporting/Cargo.toml
+++ b/codespan-reporting/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codespan-reporting"
-version = "0.1.1"
+version = "0.1.2"
 readme = "README.md"
 license = "Apache-2.0"
 authors = ["Brendan Zabarauskas <bjzaba@yahoo.com.au>"]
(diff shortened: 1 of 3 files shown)


Marwes:
So the blue styling is actually rather dark which makes it hard to see on dark terminals (at least on mine) which is why I used cyan instead as that is what ripgrep uses. rustc does not use either of those colors though if we want to use that same as it.

Marwes:
Seems like rustc special cases this to use cyan on windows https://github.com/rust-lang/rust/blob/c3fd5d0ddead5b14ce5321b06bc5bed60d0cdf2f/src/librustc_errors/emitter.rs#L1483-L1487

https://github.com/rust-lang/rust/pull/36178

brendanzab:
Ah, interesting! Are you on windows then?

Marwes:
Yep, but using Git Bash which seems to still have the issue (but it might just be my colorscheme, need to double check that actually).

PR #16, opened 2018-04-16
Make the colors more visible on Windows

### LINKED ISSUE #15: Use Cyan coloring on windows, Blue for everywhere else
@Marwes [pointed out](https://github.com/brendanzab/codespan/pull/14#issuecomment-381531763) that Blue renders quite dark by default on windows. We should special case our usages of Blue like rustc does.

Marwes:
I haven't looked at how this looks on unix so it may be that unix should only have `bold` and not `intense` enabled. Windows required `intense` since it does not understand bold.

Closes #15
 
capture

### DIFF
--- a/codespan-reporting/src/emitter.rs
+++ b/codespan-reporting/src/emitter.rs
@@ -20,7 +20,10 @@ where
     W: WriteColor,
 {
     let supports_color = writer.supports_color();
-    let line_location_color = ColorSpec::new().set_fg(Some(Color::Blue)).clone();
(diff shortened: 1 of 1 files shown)


brendanzab:
Looks good to me!

screen shot 2018-04-17 at 11 12 42 am

brendanzab:
I think Rust also bolds the `error`/`warning`/etc. part 🤔

Marwes:
Never noticed that but that seems right. I will make that change before merging.

PR #21, opened 2018-04-29
feat: Allow the files in CodeMap to be updated

Marwes:
The algorithm for reusing spans is very naive here. It only does a linear search for a place that fits (first-fit?) after checking that the previous location does not fit. Assuming that the CodeMap won't store that many files though I think this is fine, any update will be followed by a recompilation anyway which will dominate the recompile time.

Since this does not impact normal usage in any way I didn't bother creating another type.

### DIFF
--- a/codespan/Cargo.toml
+++ b/codespan/Cargo.toml
@@ -13,6 +13,7 @@ documentation = "https://docs.rs/codespan"
 failure = "0.1.1"
 serde_derive = { version = "1", optional = true }
 serde = { version = "1", optional = true }
+itertools = "0.7"
 
 [dev-dependencies]
 pretty_assertions = "0.5.0"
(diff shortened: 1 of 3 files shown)


Marwes:
> Does CodeMap::update invalidate later files?

Not sure what you mean, but it will only remove the file mentioned with the index. Since `CodeMap` does not care how large the gaps are in between the files it shouldn't affect any other files.

brendanzab:
> Seems that there is a bug with the feature gate

I thought it was more that it's on it's way to being stabilized, so the feature gate has been removed. But I might not understand how feature gates work 😛

Marwes:
Derp, you are right it gets stabilized next version. Thought I read something about it becoming stable first with the 2018 edition.

https://github.com/Aaronepower/rust/blob/master/RELEASES.md#version-1260-2018-05-10

PR #22, opened 2018-04-29
 feat: Add the codespan-lsp crate

Marwes:
This contains functions to translate codepan's types into the ones used
n the language server protocol and is an almost straight translation of the pre-codespan implementation used in gluon's language server (which is why some parts may look a bit strange).

### DIFF
--- a/codespan/src/lib.rs
+++ b/codespan/src/lib.rs
@@ -3,6 +3,7 @@
 
 #[macro_use]
 extern crate failure;
+extern crate itertools;
 #[cfg(test)]
 #[macro_use]
 extern crate pretty_assertions;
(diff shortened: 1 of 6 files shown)


Marwes:
This also contains a generalization of FileMap which I failed to split up into a separate commit. While it currently is constrained to `AsRef<str>` I'd imagine this could be extended to let `FileMap`'s contain ropes at a later stage.

codespan/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 #[macro_use]
 extern crate failure;
+extern crate itertools;
```
brendanzab: We don't seem to be using this? 🤔

PR #24, opened 2018-05-02
Implement serialization for more types

Michael-F-Bryan:
I recently tried to serialize things like `Span` and `CodeMap` using serde and found it didn't work. This PR just adds extra impls hidden behind the `serialization` feature flag and fixes a typo where `#[cfg_attr(...)]` above `extern crate serde` was using the wrong feature name.

I also added a note about optional features in the docs to aid in discoverability.

### DIFF
--- a/codespan/Cargo.toml
+++ b/codespan/Cargo.toml
@@ -20,4 +20,4 @@ pretty_assertions = "0.5.0"
 
 
 [features]
-serialization = ["serde", "serde_derive"]
+serialization = ["serde", "serde/rc", "serde_derive"]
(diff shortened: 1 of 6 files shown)


brendanzab:
Thanks! Curious what you are experimenting with using codespan for?

Michael-F-Bryan:
I like playing around with creating my own programming languages and this time I thought I'd try to create a virtual machine for some proprietary language we occasionally use at work (it's a form of BASIC for programming PLCs).

From what I've seen (and done) with `codespan` so far I think it's really awesome. I always thought the `CodeMap` abstraction from `rustc` was well done, so it's nice to see a similar thing available for the crates.io ecosystem.

I was actually about to make an issue about adding a `Diagnostics` type for collecting multiple errors/bugs/warnings. The idea being you'd give this to each pass, allowing you to accumulate all issues relevant to that pass, emitting them and stopping if the `Diagnostics` says there was at least one fatal error. I'll need to make something like that for the type checking phase, so hopefully when it's started taking shape I can send it as a PR to `codespan-diagnostics`.

PR #27, opened 2018-06-30
Add support for specifiyng an error code with a Diagnostic

garyttierney:
Small change to codespan-reporting that adds a new `code` field to the `Diagnostic` struct.  It is emitted in the same style as the Rust compiler.


brendanzab:
Nice, I like it!

PR #33, opened 2018-08-21
Add serde support to codespan-reporting

Michael-F-Bryan:
I just started putting together a "compile-test" crate for my toy C compiler and realised `Diagnostic` isn't serializable. This PR adds `#[derive(Serialize, Deserialize)]` to the various types in `codespan-reporting`.


brendanzab:
Thanks!

PR #35, opened 2018-09-12
Make emit take the writing by mutable reference

### LINKED ISSUE #34: Make codespan_repoirting::emit take a multiple reference to the writer
There doesn't seem to be any reason for it to consume the writer other than to force the user to lock the output multiple times. I just can't think of any reason for the `emit` function to take the writer argument by value and I don't know if I can say much more about this. It should be a trivial change.

boomshroom:
Fixes #34.

### DIFF
--- a/codespan-reporting/src/emitter.rs
+++ b/codespan-reporting/src/emitter.rs
@@ -15,7 +15,7 @@ impl<T: fmt::Display> fmt::Display for Pad<T> {
     }
 }
 
-pub fn emit<W>(mut writer: W, codemap: &CodeMap, diagnostic: &Diagnostic) -> io::Result<()>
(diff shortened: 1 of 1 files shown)


brendanzab:
Looks good!

brendanzab:
Do we need to update the example? Weird that travis didn't fail...

brendanzab:
Ah cool, maybe make a commit to do that?

Marwes:
WriteColor is implemented on `&mut T` which is what lets the example work and why taking the writer by value is actually more general. If we want it by reference then it should at least also have a `?Sized` bound.

https://docs.rs/termcolor/1.0.3/termcolor/trait.WriteColor.html

boomshroom:
Huh. I must have missed that and the errors I was getting was actually something else.

PR #36, opened 2018-10-02
Flexible line breaking

declanvk:
Previously the method for line breaking in `Filemap`s only considered those lines that ended with a `\n` character. This prevents interoperability with Windows, where lines are delimited with `\r\n`.

This PR adds support for both lines that end with `\r\n` and `\n`. This will consider both types in the same file to be legal, which may not be desired behavior. An alternative would be to make a check for windows vs other via a `cfg`, and then only allow a `\r\n` or `\n`, not both.

### DIFF
--- a/codespan/src/filemap.rs
+++ b/codespan/src/filemap.rs
@@ -162,7 +162,8 @@ where
         let span = ByteSpan::from_offset(start, ByteOffset::from_str(src.as_ref()));
         let lines = {
             let newline_off = ByteOffset::from_char_utf8('\n');
-            let offsets = src.as_ref()
+            let offsets = src
+                .as_ref()
                 .match_indices('\n')
                 .map(|(i, _)| ByteOffset(i as RawOffset) + newline_off);
 
@@ -226,7 +227,9 @@ where
             })
     }
 
-    /// Returns the byte offset to the start of `line`
+    /// Returns the byte offset to the start of `line`.
+    ///
+    /// Lines may be delimited with either `\n` or `\r\n`.
     pub fn line_offset(&self, index: LineIndex) -> Result<ByteOffset, LineIndexError> {
         self.lines
             .get(index.to_usize())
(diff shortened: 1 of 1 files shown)


brendanzab:
Oh thanks! Yeah, I think this seems reasonable. Do you know what Rust does in this case?

declanvk:
@brendanzab 

Based on [`str::lines`](https://doc.rust-lang.org/1.29.1/src/core/str/mod.rs.html#2846-2848) and the [`AnyLinesMap`](https://doc.rust-lang.org/1.29.1/src/core/str/mod.rs.html#1355-1364) Rust will break on any `\n` it finds, then check for the presence of a following `\r` and remove it, if it exists.

Another example would be [`BuffRead::lines`](https://doc.rust-lang.org/src/std/io/mod.rs.html#2147-2171) which has the same behavior.

So it seems that the outcome is the same, both `\n` and `\r\n` will be considered line breaks in the same string.

brendanzab:
Oh I meant more in the `Filemap` in `libsyntax_pos`.

brendanzab:
This seems like the place: https://github.com/rust-lang/rust/blob/004bc5a33cd44cd0896cae917a75488579126da5/src/libsyntax_pos/lib.rs#L1102-L1113

Not sure how they handle problems with `\r\n` 🤔

declanvk:
Ah, yeah based on some more searching the [`libsyntax_pos/lib.rs`](https://github.com/rust-lang/rust/blob/master/src/libsyntax_pos/lib.rs), I was unable to find a reference to any `\r`. Also based on this [issue](https://github.com/rust-lang/rust/issues/7946) and this [thread](https://users.rust-lang.org/t/isnt-crlf-default-on-windows/10474/2), rust itself does not accept `\r\n`.

brendanzab:
Oh, that's a neat way of handling it. But yeah, seeing as codespan is meant to be more general, perhaps we should try to handle it. Wondering how much of a performance hit we incur with this stuff, and if that would be much of a problem in practice. 🤔

declanvk:
If the performance is an issue, maybe it would be better to only allow a single line-ending type per file? That might simplify some of the logic, but it does still need to do a 2 character check for `\r\n` versus just `\n`.

Also, thanks for this library, its been a great help in some of my projects.

Marwes:
I don't think this actually changes which offsets are returned? With this implementation, each returned offset on a CRLF file is `position('\r') + '\r'.len() + '\n'.len()` but that is the exact same offset as `position('\n') + '\n'.len()` since `position('\r') + '\r'.len() == position('\n')`.

brendanzab:
Right! Would be interesting to add the `\r\n` test even without this change.

declanvk:
@brendanzab @Marwes The last commit removes the change to the line-breaking logic, but leaves the changes to the tests.

brendanzab:
Yes, that would be super useful! Would you be able to squash the commits as well?

brendanzab:
Lovely, thanks!

PR #40, opened 2018-11-12
Implement more things for FileName

Michael-F-Bryan:
This adds a couple relatively small changes to make interoperating between `FileName` and `Path`/`PathBuf` a bit easier:

- Derive `Hash` so `FileName` can be used as the key in things like `HashMap`
- `From<FileName>` for `Path` and `PathBuf` (only requires an allocation in the `Cow::Borrowed` case)
- `AsRef<Path>` (useful for things like `File::open(...)`)
-  `PartialEq<PathBuf>`/`PartialEq<Path>` for comparing a `FileName` to `Path` and `PathBuf`

### DIFF
--- a/codespan/src/filemap.rs
+++ b/codespan/src/filemap.rs
@@ -61,6 +61,27 @@ impl From<&'static str> for FileName {
     }
 }
 
+impl AsRef<Path> for FileName {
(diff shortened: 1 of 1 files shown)


brendanzab:
Thanks!

Issue #46, opened 2019-01-08
Indexing with `ByteIndex`


Laegluin:
If I'm not mistaken, `ByteIndex` uses 1-based indexing, but that is never mentioned in the docs. This should probably be documented, since I would have expected it to be 0-based.

I'd happily submit a PR with some changes :)

brendanzab:
This is because of how we don't have a separate 'file id' to differentiate files in the filemap. the '0' index is intended to denote positions that don't correspond to any file. I realize that this is confusing though! As mentioned in #52, the language-reporting crate simplifies this way of doing things greatly, by instead just going with a separate file id.

Laegluin:
Definitely makes more sense in that case, I'm not sure how I even managed to miss `ByteIndex::none` to begin with.

Laegluin:
Closed by #54.

PR #54, opened 2019-02-28
Document the indexing scheme used by `ByteIndex`

Laegluin:
Clarifies how indices work for `ByteIndex` (see #46).

### DIFF
--- a/codespan/src/index.rs
+++ b/codespan/src/index.rs
@@ -180,7 +180,30 @@ impl fmt::Display for ColumnOffset {
     }
 }
 
-/// A byte position in a source file
+/// A byte position in a source file. The index is continuous, running over all
+/// files in a `CodeMap`, starting with 1. Zero is reserved for positions that 
+/// do not point to a valid file. One byte of padding is added between each file.
+///
+/// For example, for these files:
+///
+/// - `main.rs`: `pub struct` (10 bytes)
+/// - `lib.rs`: `Span::new` (9 bytes)
+///
+/// `ByteIndex(0)` points to no file at all, `ByteIndex(2)` points at the
+/// second byte (index 1) in `main.rs` and `ByteIndex(14)` points at the third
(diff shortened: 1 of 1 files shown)


codespan/src/index.rs
```diff
@@ -180,7 +180,8 @@ impl fmt::Display for ColumnOffset {
     }
 }
 
-/// A byte position in a source file
+/// A 1-indexed byte position in a source file. Zero is reserved for positions that do not
```
Marwes: It isn't really `1-indexed` either however. Since all files reside in the same "index" it is only the first file that is `1-indexed` the others start at 316, 123745, etc

brendanzab: Agreed. I'll merge it if this is updated to be more accurate!

Laegluin:
Sorry about the delay, I must have missed your comments somehow :)
I updated the comment, if I understood correctly this should be more accurate.

codespan/src/index.rs
```diff
@@ -180,7 +180,18 @@ impl fmt::Display for ColumnOffset {
...
+/// do not point to a valid file.
+///
+/// For example, with these files:
+///
+/// - `main.rs` (10 bytes)
+/// - `lib.rs` (314 bytes)
+///
+/// `ByteIndex(0)` points to no file at all, `ByteIndex(2)` points at the
+/// second byte (index 1) in `main.rs` and `ByteIndex(14)` points at the fourth
+/// byte (index 3) in `lib.rs`.
```
Marwes: This is almost correct :). There is also another byte of padding between each file so that we can distinguish the end of one file from the start of another. (I had to double check this).

So `14` ends up pointing at index 2 of `lib.rs`

Laegluin: Fixed, hope that was the last error :)

brendanzab:
Amazing work! This is really great!

Laegluin:
Glad I could help :)

PR #55, opened 2019-03-28
Make CodeMap generic over AsRef<str>

oberien:
This should be a backwards-compatible change, as everything still defaults to `String` if no generic argument is given. It *does* allow using `&str` instead of `String` for file contents though.

### DIFF
--- a/codespan/src/codemap.rs
+++ b/codespan/src/codemap.rs
@@ -7,16 +7,16 @@ use itertools::Itertools;
 use filemap::{FileMap, FileName};
 use index::{ByteIndex, ByteOffset, RawIndex};
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 #[cfg_attr(feature = "serialization", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "memory_usage", derive(HeapSizeOf))]
-pub struct CodeMap {
-    files: Vec<Arc<FileMap>>,
+pub struct CodeMap<S: AsRef<str> = String> {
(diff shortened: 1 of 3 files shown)


codespan/src/codemap.rs
```diff
@@ -7,16 +7,16 @@ use itertools::Itertools;
 use filemap::{FileMap, FileName};
 use index::{ByteIndex, ByteOffset, RawIndex};
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 #[cfg_attr(feature = "serialization", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "memory_usage", derive(HeapSizeOf))]
-pub struct CodeMap {
-    files: Vec<Arc<FileMap>>,
+pub struct CodeMap<S: AsRef<str> = String> {
```
Marwes: Specifying bounds on type definitions is generally a bad idea as it forces it to specified in all uses of the type regardless of whether it is needed or not (as you have seen from the explicit `Default` impl).

So this bound can just be removed, the `Default` impl derived and the bound on the `impl` block only needs to be for a few specific methods.

oberien: I need to keep the generics on the impl-blocks, but not the trait-bounds (for most methods) to still allow things other than `String` to be added to the CodeMap. Other than that, it is probably for the better if I change it.

oberien: Actually all methods except `new` (given the custom Default implementation) and `iter` require the `AsRef<str>` bound anyway.

brendanzab:
Thanks! :)

PR #56, opened 2019-03-29
Add `emit_single`, which emits a Diagnostic given a single FileMap

oberien:
My program performs a sort-of recursive-descend approach to files. As such I don't have all files loaded at the same time. Thus, it's hard for me to use a `CodeMap`. Instead, I'd like to be able to emit diagnostics just given a single `FileMap`.

This PR adds a `emit_single` function, which can be used to emit diagnostics given a `FileMap` instead of a `CodeMap`.


ebkalderon:
Since `CodeMap` and `FileMap` were replaced by a single `Files` type in version 0.4.0, is this pull request still relevant?

brendanzab:
Yeah, I think this can be closed now.

PR #57, opened 2019-04-29
feat: Add new_unchecked as a const function

Marwes:
Needed a way to construct a span in a const context.


brendanzab:
Seems like we need to increase the minimum suported version on travis

PR #58, opened 2019-05-11
Constify some index methods

etaoins:
This was originally motivated by wanting this for tests:

```rust
pub const EMPTY_SPAN: Span = Span::new_unchecked(ByteIndex::none(), ByteIndex::none());
```

However it seems a few other methods can be safely converted to `const` at the same time.


brendanzab:
Thanks!

PR #59, opened 2019-05-12
Add Arret to projects

etaoins:
Since etaoins/arret@7cd236e it uses `codespan` and `codespan-reporting` for all of its source locations and diagnostics.


brendanzab:
Lovely!

PR #74, opened 2019-06-03
Add additional notes to marked source section

brendanzab:
Adds notes below the marked source section:

```
   ┌╴ <test>:2:9
   │
 2 │ (+ test "")
   │         ^^ Expected integer but got string
   │
   = perhaps you meant a number like `1`?
```

### DIFF
--- a/codespan-reporting/src/diagnostic.rs
+++ b/codespan-reporting/src/diagnostic.rs
@@ -59,14 +59,17 @@ pub struct Label {
     pub span: ByteSpan,
     /// A message to provide some additional information for the underlined code.
     pub message: String,
+    /// Notes providing more information under the source code snippet.
+    pub notes: Vec<String>,
 }
 
 impl Label {
     /// Create a new label.
-    pub fn new(span: ByteSpan, message: impl Into<String>) -> Label {
+    pub fn new(span: ByteSpan, message: impl Into<String>, notes: impl Into<Vec<String>>) -> Label {
         Label {
             span,
             message: message.into(),
+            notes: notes.into(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


codespan-reporting/src/diagnostic.rs
```diff
@@ -59,14 +59,17 @@ pub struct Label {
     pub span: ByteSpan,
     /// A message to provide some additional information for the underlined code.
     pub message: String,
+    /// Notes providing more information under the source code snippet.
+    pub notes: Vec<String>,
```
brendanzab: Not sure if this should be in the `Label` or the `Diagnostic`. It could make it trickier if we ever wanted to try to group underlines into a single source section like rustc does. 🤔

Marwes: > It could make it trickier if we ever wanted to try to group underlines into a single source section like rustc does. 🤔

I would agree with that, if one wants labels separately it probably makes more sense to have them in different diagnostics as well.

I think of each label as pointing out a specific location in the source, having additional "notes" which are related to the label but not really related to the position (then they would be part of the message instead) does not seem as useful.

brendanzab: Maybe we could have something like this?

```
   ┌── test:2:9 ───
   │
 2 │ (+ test "")
   │         ^^ expected `Int` but got `String`
   │
   │     ┌── prims:102:1 ───────────────────────────┐
   │     │                                          │
   │ 102 │ (define + (x y) ...)                     │
   │     │ -------------------- `+` defined here    │
   │     │                                          │
   │     └──────────────────────────────────────────┘
   │
   = perhaps you meant a number like `1`?
```

brendanzab: Another option:

```
   ┌── test:2:9 ───
   │
 2 │ (+ test "")
   │         ^^ expected `Int` but got `String`
   │
   = perhaps you meant a number like `1`?
   = blah blah blah

        ┌── prims:102:1 ───
        │
    102 │ (define + (x y) ...)
        │ -------------------- `+` defined here
        │
```

PR #75, opened 2019-06-06
Break up emitter module

brendanzab:
The emitter module is getting quite huge! Splitting things up a little now.


codespan-reporting/src/emitter/views/border_left.rs
```diff
@@ -0,0 +1,23 @@
+use std::io;
+use termcolor::{ColorSpec, WriteColor};
+
+use crate::emitter::Config;
+
+/// The left-hand border of a source line.
+pub struct BorderLeft {}
+
+impl<'a> BorderLeft {
```
Schyrsivochter: Why is there a lifetime parameter here?

brendanzab: oh, that is a mistake - weird that rustc didn't complain

brendanzab: Good catch - it was there in a few of them - copy paste issue I think

Issue #77, opened 2019-06-06
Improve multiline errors


brendanzab:
As can be seen in the emitter example the error for multiline errors could be improved.

Currently we render:

```
   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │ fizz₁ num = case (mod num 5) (mod num 3) of
 5 │     0 0 => "FizzBuzz"
 6 │     0 _ => "Fizz"
 7 │     _ 0 => "Buzz"
 8 │     _ _ => num
   │             -------------- `case` clauses have incompatible types
   │
```

Instead, it might be nice if we followed rustc's lead, and tried to mark around the region instead:

```
   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────╯
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ ╰──────────────╯ `case` clauses have incompatible types
   │
```

Eventually we could try to combine errors like so:

```
error[E0308]: `case` clauses have incompatible types

   ┌── FizzBuzz.fun:8:12 ───
   │
 3 │   fizz₁ : Nat → String
   │                 ------ expected type `String` found here
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────╯
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ │            ^^^ expected `String`, found `Nat`
   │ ╰──────────────╯ `case` clauses have incompatible types
   │
   = expected type `String`
        found type `Nat`
```

brendanzab:
Rustc tends to handle overlapping multiline errors cleverly too:

<img width="1137" alt="Screen Shot 2019-06-04 at 2 14 32 pm" src="https://user-images.githubusercontent.com/695077/59042383-af22ca00-88bd-11e9-9ed0-439586b6f71e.png">

brendanzab:
Might be worth studying how https://github.com/rust-lang/rust/tree/961a9d6e97a37ff1bb33578721dfa8687054c7b8/src/librustc_errors does this

brendanzab:
Here's another idea for primary spans:

```
   ┌── FizzBuzz.fun:4:13 ───
[...]

PR #98, opened 2019-06-23
Implement snakey underlines!

brendanzab:
Working towards #77, this removes some of the useless space we were printing out at the end of empty lines, and fixes the length of multiline underlines. This means that instead of:

```
   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │ fizz₁ num = case (mod num 5) (mod num 3) of
 5 │     0 0 => "FizzBuzz"
 6 │     0 _ => "Fizz"
 7 │     _ 0 => "Buzz"
 8 │     _ _ => num
   │             -------------- `case` clauses have incompatible types
   │
```

We now print:

```
   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────'
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ ╰──────────────' `case` clauses have incompatible types
   │
```

For cases when we have empty space at the beginning, we print:

```text
    ┌── FizzBuzz.fun:11:5 ───
    │
 11 │ ╭     case (mod num 5) (mod num 3) of
 12 │ │         0 0 => "FizzBuzz"
 13 │ │         0 _ => "Fizz"
 14 │ │         _ 0 => "Buzz"
 15 │ │         _ _ => num
    │ ╰──────────────────' `case` clauses have incompatible types
    │
```

🐍🐍 🎉


codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color.snap
```diff
@@ -1,71 +1,72 @@
...
- 4 │ fizz₁ num = case (mod num 5) (mod num 3) of
- 5 │     0 0 => "FizzBuzz"
- 6 │     0 _ => "Fizz"
- 7 │     _ 0 => "Buzz"
- 8 │     _ _ => num
-   │             -------------- `case` clauses have incompatible types
-   │ 
+   │
+ 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
+   │ ╭─────────────-
```
Marwes: Shouldn't `-` be an arrow of sorts? It is rather ambigous wheter this points up or down (or at all)

brendanzab: Yeah, I mentioned this here: https://github.com/brendanzab/codespan/issues/77#issuecomment-504724596

I think rustc gets away with this because it does:

```
___-
```

brendanzab: One alternative is using another box drawing thing for secondary underlines:

```
─────────┘
```
or
```
─────────╯
```

brendanzab: Open to ideas though!

brendanzab: Another idea:

```
─‾
```

Using OVERLINE (U+203E) - might be less well supported by mono fonts though :/
```

Comment (Marwes) [reply codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color.snap]: Using another box drawing makes it clear at least! I am well versed in unicode drawing so I don't have a better suggestion.

Overline looks good as well but if it can't be relied on to be there it doesn't seem like a good idea.

Comment (brendanzab) [reply codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color.snap]: Here's what overbar looks like alongside the carets we use for primary underlines:

```
foo <$> bar
‾‾‾ ^^^ ‾‾‾
```
doesn't look too bad I guess 🤔

Comment (brendanzab) [reply codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color.snap]: Another option using `'`:

```
  fizz₁ num = case (mod num 5) (mod num 3) of
╭─────────────'
```
```
foo <$> bar
''' ^^^ '''
```
And another using `"`
```
  fizz₁ num = case (mod num 5) (mod num 3) of
╭─────────────"
```
```
foo <$> bar
""" ^^^ """
```

Comment (brendanzab) [reply codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color.snap]: Oooh, perhaps we could combine `-` and `'` for secondary labels:

```
  fizz₁ num = case (mod num 5) (mod num 3) of
╭─────────────'
```
```
foo <$> bar
--- ^^^ --- a label about `bar`
```

brendanzab: Tried this out in https://github.com/brendanzab/codespan/pull/98/commits/46cf1255c38c09325bd1f40cc6b7119be70dba0c

PR #99, opened 2019-06-23
Don’t highlight source with color

brendanzab:
This used to be necessary before we had the snakey underlines (#98), but now it's less useful, especially in multiline sections. It will also make it harder to implement overlapping error messages like in rustc (#100).

Before:

Screen Shot 2019-06-24 at 9 23 20 am

After:

Screen Shot 2019-06-24 at 9 23 07 am

### DIFF
--- a/codespan-reporting/src/term/views/diagnostic.rs
+++ b/codespan-reporting/src/term/views/diagnostic.rs
@@ -2,8 +2,8 @@ use codespan::{Files, Location};
 use std::io;
 use termcolor::WriteColor;
 
-use crate::term::Config;
 use crate::diagnostic::Diagnostic;
+use crate::term::Config;
 
 use super::{Header, Locus, NewLine, SourceSnippet};
 
--- a/codespan-reporting/src/term/views/header.rs
+++ b/codespan-reporting/src/term/views/header.rs
@@ -1,8 +1,8 @@
 use std::io;
 use termcolor::WriteColor;
 
-use crate::term::Config;
 use crate::diagnostic::{Diagnostic, Severity};
+use crate::term::Config;
 
 use super::NewLine;
 
--- a/codespan-reporting/src/term/views/source_snippet/mod.rs
+++ b/codespan-reporting/src/term/views/source_snippet/mod.rs
@@ -89,7 +89,6 @@ impl<'a> SourceSnippet<'a> {
         let end_line_span = self.line_span(end.line).expect("end_line_span");
         let is_multiline = start.line != end.line;
 
-        let label_style = self.mark_style.label_style(config);
         // Use the length of the last line number as the gutter padding
         let gutter_padding = format!("{}", end.line.number()).len();
         // Cache the tabs we'll be using to pad the source strings.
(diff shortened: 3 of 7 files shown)


brendanzab:
I'm open to thoughts and feedback on this.

lpil:
I think this is slightly clearer with colour, as it is now.

Marwes:
I'd say keep the colors. Only in the multiline case is arguably a bit clearer since coloring ends up making the code blend in rather than stand out. Apart from that the color makes it clearer.

brendanzab:
Oh interesting! Thanks for the feedback. Would you still like it in case of things like:

lpil:
Would it be possible to get an image with colours? I'm finding it hard to imagine 😅

brendanzab:
Ugh, I guess I'd have to implement the grouped labels stuff then. 😅 - that's non-trivial 😬

lpil:
What does it do without this PR?

brendanzab:
I would have to figure out how to group labels together (see this issue #100), and manage the spans of highlighted source as well. This makes things even trickier for me to implement. Imagine this: https://github.com/rust-lang/rust/blob/961a9d6e97a37ff1bb33578721dfa8687054c7b8/src/librustc_errors/emitter.rs but with source highlighting as well!

brendanzab:
Perhaps this should be part of a PR for the grouping... that way if I take something away at least I'll put something better in its place.

brendanzab:
I mean, at the moment I consider grouping labels to be of a higher priority than coloured source spans. This collapses something like:

```
error[E0308]: `case` clauses have incompatible types

   ┌── FizzBuzz.fun:8:12 ───
   │
 8 │     _ _ => num
   │            ^^^ expected `String`, found `Nat`
   │
   = expected type `String`
        found type `Nat`

   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────'
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ ╰──────────────' `case` clauses have incompatible types
   │

   ┌── FizzBuzz.fun:3:15 ───
   │
 3 │ fizz₁ : Nat → String
   │               ------ expected type `String` found here
   │
```

Into something much smaller:

```
error[E0308]: `case` clauses have incompatible types

   ┌── FizzBuzz.fun:8:12 ───
   │
 3 │   fizz₁ : Nat → String
   │                 ------ expected type `String` found here
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────'
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ │            ^^^ expected `String`, found `Nat`
   │ ╰──────────────' `case` clauses have incompatible types
   │
   = expected type `String`
        found type `Nat`
```

brendanzab:
The trickiness comes in handling all the edge cases of overlapping labels and underlines.

Marwes:
Maybe we could have a rule that a label colors the source unless there is another label whose span is contained in the first label.

Not sure what to do if they partially overlap however.

brendanzab:
I'm guessing the primary spans would take priority

Marwes:
Yeah, but you could also have two secondary spans that partially overlap.

brendanzab:
I guess you could merge the highlighted bits together in that case

brendanzab:
One idea might be to only highlight the source code of primary spans. 🤔

lpil:
That sounds good

brendanzab:
Ok, I'm going to close this for now, thanks again for the feedback! ❤️

PR #103, opened 2019-06-24
Group character config into a single struct

brendanzab:
This makes managing character sets more manageable.

### DIFF
--- a/codespan-reporting/src/term/config.rs
+++ b/codespan-reporting/src/term/config.rs
@@ -13,49 +13,8 @@ pub struct Config {
     pub tab_width: usize,
     /// Styles to use when rendering the diagnostic.
     pub styles: Styles,
-
-    /// The character to use for the top-left border of the source.
-    /// Defaults to: `'┌'`.
-    pub source_border_top_left_char: char,
-    /// The character to use for the top border of the source.
-    /// Defaults to: `'─'`.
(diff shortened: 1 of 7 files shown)


codespan-reporting/tests/snapshots/fizz_buzz__rich_no_color_ascii.snap
```diff
@@ -0,0 +1,72 @@
...
+         found type `Nat`
+
+    --- FizzBuzz.fun:11:5 ---
+    |
+ 11 |  ____ case (mod num 5) (mod num 3) of
+ 12 | |         0 0 => "FizzBuzz"
+ 13 | |         0 _ => "Fizz"
+ 14 | |         _ 0 => "Buzz"
+ 15 | |         _ _ => num
+    | |__________________- `case` clauses have incompatible types
```
brendanzab: I'm unsure about:

```

 11 |  ____ case (mod num 5) (mod num 3) of
 12 | |         0 0 => "FizzBuzz"
```

Should I just go with:

```
 11 |  /    case (mod num 5) (mod num 3) of
 12 | |         0 0 => "FizzBuzz"
```

Like in Rustc? It does look better with color FWIW

brendanzab:
Think I might merge the character struct change and leave the ASCII thing for another PR

PR #118, opened 2019-08-23
Implement Default for Span

### LINKED ISSUE #117: Implement or derive Default for Span
Prior to the release of `codespan` 0.4.0, the `codespan::Span` type used to derive `Default`. I personally found this quite handy when hacking on ASTs where I did not care about the span information, and I could conveniently specify  `Default::default()` everywhere a `Span` was expected without having to import it.

In particular, I also happen to find the `Default` implementation handy when writing macros that construct ASTs, since they eliminate the need to depend on `codespan` and import the type. So if a downstream crate were to import the macro, it would not require the user to depend on `codespan`, or for the crate author to re-export `codespan::Span` for the macro to use.

Are there any technical reasons to not derive `Default` on `Span`? If there aren't, it would be a nice touch in the ergonomics department.

ebkalderon:
### Changed

* Implement `std::default::Default` for `Span` to aid API ergonomics.

Closes #117.

### DIFF
--- a/codespan/src/span.rs
+++ b/codespan/src/span.rs
@@ -84,6 +84,12 @@ impl Span {
     }
 }
 
+impl Default for Span {
+    fn default() -> Span {
+        Span::initial()
+    }
+}
+
 impl fmt::Display for Span {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(


codespan/src/span.rs
```diff
@@ -5,7 +5,7 @@ use std::ops::Range;
 
 use crate::ByteIndex;
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
brendanzab: Perhaps we could implement `Default` using `Span::intitial` in an explicit `impl`? I'm a tad uncomfortable leaving this up to a derive. 🤔

ebkalderon: Sure, that would probably be better. I'll modify the PR. Thanks for the suggestion!

ebkalderon: @brendanzab Committed, updated, and ready for re-review!

brendanzab: Thanks!

Issue #100, opened 2019-06-24
Group labels originating from the same file into a single snippet


brendanzab:
As mentioned in #77, Rustc does a clever job of grouping together labels:

![](https://user-images.githubusercontent.com/695077/59042383-af22ca00-88bd-11e9-9ed0-439586b6f71e.png)

It would be super neat if we could do something similar!

At the moment our diagnostics are all spread out:

```
error[E0308]: `case` clauses have incompatible types

   ┌── FizzBuzz.fun:8:12 ───
   │
 8 │     _ _ => num
   │            ^^^ expected `String`, found `Nat`
   │
   = expected type `String`
        found type `Nat`

   ┌── FizzBuzz.fun:4:13 ───
   │
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────'
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ ╰──────────────' `case` clauses have incompatible types
   │

   ┌── FizzBuzz.fun:3:15 ───
   │
 3 │ fizz₁ : Nat → String
   │               ------ expected type `String` found here
   │
```

It would be great if we could collapse things down, for example:

```
error[E0308]: `case` clauses have incompatible types

   ┌── FizzBuzz.fun:8:12 ───
   │
 3 │   fizz₁ : Nat → String
   │                 ------ expected type `String` found here
 4 │   fizz₁ num = case (mod num 5) (mod num 3) of
   │ ╭─────────────'
 5 │ │     0 0 => "FizzBuzz"
 6 │ │     0 _ => "Fizz"
 7 │ │     _ 0 => "Buzz"
 8 │ │     _ _ => num
   │ │            ^^^ expected `String`, found `Nat`
   │ ╰──────────────' `case` clauses have incompatible types
   │
   = expected type `String`
        found type `Nat`
```

Or in the case of that rustc error:

```
error[E0308]: match arms have incompatible types

    ┌─ codespan/src/file.rs:102:34
    :
99  │  ╭──────── match line_index.compare(self.last_line_index()) {
100 │  │             Ordering::Less => Ok(self.line_starts()[line_index.to_usize()]),
    │  │                               --------------------------------------------- this is found to be of type `std::result::Result<index::ByteIndex, file::LineIndexOutOfBoundsError>`
[...]

PR #121, opened 2019-08-28
Group snippets by file

brendanzab:
This is a step towards #100. Instead of printing out individual snippets for each labelled span, we group them together on a per-file basis.


brendanzab:
This is what it looks like:

Screen Shot 2019-08-29 at 6 35 01 pm

PR #127, opened 2019-10-03
feat: Allow passing trait objects as WriteColor

Marwes:


brendanzab:
Ooh, nice! Wondering is we should just use trait objects in the API to reduce code size/compilation time 🤔

codespan-reporting/tests/term.rs
```diff
@@ -1,11 +1,19 @@
...
 
 mod support;
 
 use self::support::TestData;
 
+fn test_config() -> Config {
+    Config {
+        // Always use blue so tests are consistent across platforms
+        styles: Styles::with_blue(Color::Blue),
+        ..Config::default()
```
brendanzab: Great idea! 👍

Marwes:
Maybe, but that could arguably be done by the caller (just pass `&mut dyn WriteColor`).

Issue #138, opened 2019-12-27
Make the Label an optional part of the Diagnostic


prybicki:
Sometimes there's no specific place in the source file to point - consider a program in C which does not have main() function.

jyn514:
@prybicki the way I currently handle this is by pointing to the end of the file. YMMV, though.

brendanzab:
Been experimenting with a new API that will let you do something like:

```rust
Diagnostic::error("oops")
```

This will emit something like:

```
error: oops
```

PR #144, opened 2020-01-04
Make `FileId` non-zero

etaoins:
When porting from `codespan` 0.3 Arret I needed to track the `FileId` separately from the `Span` now that the file isn't implied in the span itself. Having a simple struct like this works for 90% of the cases:

```rust
struct Span {
  file_id: codespan::FileId,
  codespan_span: codespan::Span,
}
```

However, this requires that every time we parse any source we require a `FileId` which requires a `codespan::Files` to create. This is overkill in a couple of cases:

1. Unit tests that just need to parse simple Arret strings as a shorthand for building our AST

2. Syntax highlighting in the REPL

The compromise in etaoins/arret#160 is to use `Option<FileId>`. I'm not 100% happy with that as the ultimate solution but it was the easiest path to take during porting. However, this makes the size of the above struct 13 bytes which will typically be rounded up to 16 bytes for alignment and padding.

We can use `std::num::NonZeroU32` instead and offset all of our indexes by 1. This allows the Rust compiler to use 0 to represent the `None` case and bring the size down to 12 bytes. Because the struct only needs an alignment of 4 bytes this is more likely to stay as 12 bytes in most contexts.

A nicer solution would be if Rust had e.g. `NonMaxU32` but it appears the Rust team instead wants to use const generics so arbitrary niche values can be specified.

### DIFF
--- a/codespan/src/file.rs
+++ b/codespan/src/file.rs
@@ -67,8 +67,15 @@ impl fmt::Display for SpanOutOfBoundsError {
 /// A handle that points to a file in the database.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serialization", derive(Deserialize, Serialize))]
-#[cfg_attr(feature = "memory_usage", derive(heapsize_derive::HeapSizeOf))]
 pub struct FileId(NonZeroU32);
+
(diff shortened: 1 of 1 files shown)


brendanzab:
Awesome, I've been thinking about this too. This might make #138 easier too. 🤔

etaoins:
@brendanzab 

I think so. From what I understand the compiler heuristics are pretty clever. Having a `FileId` anywhere in a struct should make an `Option` of that struct "free" from a memory layout perspective.

PR #161, opened 2020-02-20
Diagnostic API cleanups

### LINKED ISSUE #20: Builder DSL for diagnostics and labels looks clunky after autoformatting
Our builder DSL for diagnostic messages doesn't really look the nicest when passed through rustfmt. An example:

https://github.com/brendanzab/pikelet/blob/4f2327f0dfa9ebdb6b48ff3f01574db6f3421775/src/semantics/errors.rs

Maybe we could switch around how we pass arguments and call methods so that it reads a little nicer?

### LINKED ISSUE #138: Make the Label an optional part of the Diagnostic

### LINKED ISSUE #159: Deprecate the diagnostic conversion part of codespan-lsp
Having done some work with diagnostic formatting, I'm feeling more that language implementers should be creating intermediate enumerations of for their diagnostic messages, then sending those to various front-ends - be they a CLI, or a language server, etc.

This would help to make it easier to craft messages for those specific front-end, and save us the pain of trying to convert a common diagnostic format between front-ends. At the moment they don't fully match up, and this means we have to do some messy hacks in order to make it work .

brendanzab:
This is a breaking change!

- Allows for optional titles.
- Labels are no longer required for diagnostic printing. (Closes #138)
- More than one primary label can be supplied.
- Makes the builder API much nicer to work with! (Closes #20)
- Remove diagnostic conversions in codespan-lsp. (Closes #159)


brendanzab:
I think I'm getting close to the point where I can snip the link between `codespan` and `codespan-reporting`! I think this can also be added to this PR, but I'll do some experimentation!

codespan-reporting/src/term.rs
```diff
@@ -21,12 +21,56 @@ pub fn emit<Source: AsRef<str>>(
...
-        DisplayStyle::Rich => RichDiagnostic::new(diagnostic).emit(files, writer, config),
-        DisplayStyle::Short => ShortDiagnostic::new(diagnostic).emit(files, writer, config),
+    for label in &diagnostic.labels {
+        let mark_style = match label.style {
+            LabelStyle::Primary => MarkStyle::Primary(diagnostic.severity),
+            LabelStyle::Secondary => MarkStyle::Secondary,
+        };
+        label_groups
+            .entry(label.file_id)
+            .or_insert(vec![])
```
Marwes:
```suggestion
            .or_default()
```

brendanzab: Tbh I've always been a tad suspicious of using `Default` for stuff like monoidal identities… kind of would prefer to leave it explicit. 😅

codespan-reporting/tests/snapshots/term__empty__no_color.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: codespan-reporting/tests/term.rs
+expression: TEST_DATA.emit_no_color(&TEST_CONFIG)
+---
+bug:
```
brendanzab: Not sure about this output - seems rather challenging to do anything useful if there is no title? 😬

jyn514: I think it would be a good idea to require at least a title in each diagnostic.

codespan-reporting/src/term.rs
Marwes: The main nit is that `or_insert(vec![])` creates an (empty) `Vec` regardless of if the id exists. `or_insert_with(Vec::new)` is equivalent to `or_default` but a bit longer.

I'd agree if the value were generic, but since the value is a concret `Vec` I don't find `or_default` any surprising :man_shrugging:

brendanzab: Hmm, I guess I thought that seeing as `Vec::new()` doesn't allocate, it would be fine? But I do see your point.

Marwes: Yeah, it is very unlikely to matter performance wise. It is more of a clippy warning I'd say.

codespan-reporting/tests/snapshots/term__empty__no_color.snap
brendanzab: Yeah... unsure how to prevent people from doing this though - I guess people will see that it renders kind of weirdly and avoid doing it?

codespan-reporting/src/term/views/diagnostic.rs
```diff
@@ -87,16 +160,25 @@ where
...
-            .expect("origin");
-        let location = files
-            .location(label.file_id, label.range.start)
-            .expect("location");
-        Locus::new(origin, location).emit(writer, config)?;
-        write!(writer, ": ")?;
-        Header::new(self.diagnostic).emit(writer, config)?;
+        let mut primary_labels = 0;
+
+        for label in &self.diagnostic.labels {
```
Marwes:
```suggestion
        for label in self.diagnostic.labels.iter().filter(|label| label.style == LabelStyle::Primary) {
```

codespan-reporting/tests/snapshots/term__multifile__rich_no_color.snap
```diff
@@ -22,16 +22,15 @@ warning: unused parameter pattern: `n₂`
...
     ┌── Data/Nat.fun:11:1 ───
     │
  11 │ _+_ : Nat → Nat → Nat
     │ --------------------- based on the definition of `_+_`
     │
+    ┌── Test.fun:4:11 ───
+    │
+  4 │ _ = 123 + "hello"
+    │           ^^^^^^^ expected `Nat`, found `String`
+    │
```
Marwes: The order changing here doesn't seem right? Isn't this the primary error and the `Data/Nat.fun` part a hint?

brendanzab: Yeah - this is an interesting design question. My long term plan was to order the labels in a single diagnostic first by the origin, and then by the line number, as opposed to prioritising the primary diagnostics. So the label messages should be readable no matter what order they are printed in. I realise this test doesn't demonstrate this well though. 

If folks want these diagnostics in order, then they could put them in separate diagnostics, say, with a follow-up 'note' diagnostic. Not sure though, willing to hear your thoughts on this!

brendanzab: One potential solution of this is to make the `labels` section richer... allowing users to be more precise about whether they want to provide an ordered sequence, or an out of order set of labels, grouped by file id.

brendanzab:
I'm going to merge this now, but folks are welcome to provide input/suggestions with regards to finer-grained sequencing of source labels, as discussed in https://github.com/brendanzab/codespan/pull/161#discussion_r385992052

codespan-reporting/tests/snapshots/term__multifile__rich_no_color.snap
Marwes: Not sure what the best way is, but given a primary error and some help I would likely create the primary message first followed by the note and they would be created in conjuction. So I would expect that they were displayed in that way as well 🤷‍♂

brendanzab: I guess what I'm preparing for is a world where we merge the labels (#100), and you get a diagnostic that reads like a bit of source code. Kind of like:

```

    ┌─ codespan/src/file.rs:102:34
    :
99  │  ╭──────── match line_index.compare(self.last_line_index()) {
100 │  │             Ordering::Less => Ok(self.line_starts()[line_index.to_usize()]),
    │  │                               --------------------------------------------- this is found to be of type `std::result::Result<index::ByteIndex, file::LineIndexOutOfBoundsError>`
101 │  │             Ordering::Equal => Ok(self.source_span().end()),
    │  │                                ---------------------------- this is found to be of type `std::result::Result<index::ByteIndex, file::LineIndexOutOfBoundsError>`
102 │  │             Ordering::Greater => LineIndexOutOfBoundsError {
    │ ╭│──────────────────────────────────^
103 │ ││                 given: line_index,
104 │ ││                 max: self.last_line_index(),
105 │ ││             },
    │ ╰│─────────────^ expected enum `std::result::Result`, found struct `file::LineIndexOutOfBoundsError`
106 │  │         }
    │  ╰─────────' `match` arms have incompatible types
    :
    = note: expected type `std::result::Result<index::ByteIndex, file::LineIndexOutOfBoundsError>`
               found type `file::LineIndexOutOfBoundsError`
```
Not sure how to support both label ordering mattering, and it not at the same time 🤔

Marwes: It is ok that they appear in a different order if the labels are merged, however in this case the diagnostics were from different files so they wouldn't be possible to merge. In that case the order is more important.

Issue #157, opened 2020-02-08
Allow setting custom line numbers


jyn514:
This is a bit of an edge case ... In C there is a `#line` macro that allows setting the current line number, it looks like this:

```
void f();
int main() {
    #line 1
    return f();
}
<stdin>:1:12: error: returning 'void' from a function with incompatible result type
      'int'
```

Note how the error message says the error occurs on line 1, not line 4.

Currently, it looks like the lines in a file [are calculated once](https://docs.rs/codespan/0.7.0/src/codespan/file.rs.html#318) and can't be modified. Would it be possible to make the line numbers configurable?

This has the unfortunate complication that there's no longer a 1-1 mapping between the line numbers displayed to the user and the lines in a file: in my example above, there are 2 lines with line number 1. However, this corresponds pretty closely to the current difference between `LineIndex` (internal) and `LineNumber` (user-facing) so I don't forsee it being a _giant_ deal.

brendanzab:
Ahh wow! That is a pretty challenging use case? Any idea how we might be able to support this?

jyn514:
It looks like currently we store a Vec<ByteIndex>, where finding the start of the line is found by indexing into the vec using the line index. I think this is fine: the frontend wants to know the span for the physical line, not for the logical line.

The only thing that needs to change is `LineIndex::number()`. Currently that just returns `self.0 + 1`. Maybe instead we could store a `BTreeMap` of physical lines -> offsets, where the offset records the difference between the physical line and the logical line (line changes within a column is way more than I need and would be harder, I don't think we need to support it). For example, given the following code:

```c
// blank
#line 1
int main() {}
#line 10
int f() {}
```

the `offsets` would look like `{ 2: -1, 4: 6 }`. Then to find the logical line for a physical line, you look up the first entry before and add that value to the physical line. So to get the logical line for physical line 6 (`int f() {}`), we'd find the first entry before was `4: 6` and return `6 + 6 = 12` for the logical line.

Let me know if that sounds doable.

jyn514:
Oh forgot to mention: the API would look something like `Files::add_line_offset(&mut self, file: FileId, byte_index: ByteIndex)` and we would look up the appropriate line number for that index. I have no opinions on what should happen if it's called multiple times for the same line, maybe just overwrite the existing offset?

brendanzab:
@Marwes would this help you out with macro support on Gluon?

brendanzab:
Wondering if we could have a `Files::line_number` method:

[...]

PR #164, opened 2020-02-24
Create Files trait

### LINKED ISSUE #79: Custom databases (Salsa support)
It would be nice to be able to define custom `Files`-style databases depending on the needs of a project. 

- People might want to use something like [Salsa](https://github.com/salsa-rs/salsa/) to manage their source files.
- When starting off building a language, a 'SingleFile' database (that only took a single file) might be a better solution than #56.
- Others might want to use a more compact span representation than the one used currently. For example compressing it, as is done in rustc.

The [language-reporting](https://github.com/wycats/language-reporting) fork defines [`ReportingSpan` and `ReportingFiles`](https://github.com/wycats/language-reporting/blob/b021c87e0d4916b5f32756151bf215c220eee52d/src/span.rs#L18-L42) traits which might be useful for inspiration.

brendanzab:
This will allow clients of `codespan-reporting` to implement their own file databases. This is important if we want to allow users to work with frameworks like Salsa, or have domain-specific concerns with regard to file handling.

Closes #79.

### DIFF
--- a/codespan-reporting/src/diagnostic.rs
+++ b/codespan-reporting/src/diagnostic.rs
@@ -1,9 +1,9 @@
 //! Diagnostic data structures.
 
-use codespan::{FileId, Span};
 #[cfg(feature = "serialization")]
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
+use std::ops::Range;
 
 /// A severity level for diagnostic messages.
 ///
@@ -54,22 +54,26 @@ impl PartialOrd for Severity {
 /// A label describing an underlined region of code associated with a diagnostic.
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
-pub struct Label {
+pub struct Label<FileId> {
     /// The file that we are labelling.
     pub file_id: FileId,
-    /// The span we are going to include in the final snippet.
-    pub span: Span,
+    /// The range we are going to include in the final snippet.
+    pub range: Range<usize>,
     /// A message to provide some additional information for the underlined
     /// code. These should not include line breaks.
     pub message: String,
 }
 
-impl Label {
+impl<FileId> Label<FileId> {
     /// Create a new label.
-    pub fn new(file_id: FileId, span: impl Into<Span>, message: impl Into<String>) -> Label {
+    pub fn new(
+        file_id: FileId,
+        range: impl Into<Range<usize>>,
+        message: impl Into<String>,
+    ) -> Label<FileId> {
         Label {
             file_id,
-            span: span.into(),
+            range: range.into(),
             message: message.into(),
         }
     }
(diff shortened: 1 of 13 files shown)


brendanzab:
I'm not sure if this API would support people who wanted custom line-numbers - see #157.

codespan-reporting/src/diagnostic.rs
```diff
@@ -113,45 +121,148 @@ impl Diagnostic {
...
+    type Origin: std::fmt::Display;
+    type LineSource: AsRef<str>;
+
+    /// The origin of a file.
+    fn origin(&self, id: Self::FileId) -> Option<Self::Origin>;
+
+    /// The line at the given index.
+    fn line(&self, id: Self::FileId, line_index: usize) -> Option<Line<Self::LineSource>>;
+
+    /// The index of the line at the given line index.
```
jyn514:
```suggestion
    /// The index of the line at the given byte index.
```

codespan-reporting/src/diagnostic.rs
```diff
@@ -113,45 +121,148 @@ impl Diagnostic {
...
+    fn line_index(&self, id: Self::FileId, byte_index: usize) -> Option<usize>;
+
+    /// The location of the given byte index.
```
jyn514:
```suggestion
    /// A user-facing location representing the given byte index.
```

codespan-reporting/src/diagnostic.rs
```diff
@@ -113,45 +121,148 @@ impl Diagnostic {
...
+    fn location(&self, id: Self::FileId, byte_index: usize) -> Option<Location> {
+        let line_index = self.line_index(id, byte_index)?;
+        let column_index = self.line(id, line_index)?.column_index(byte_index)?;
+
+        Some(Location {
+            line: line_index,
+            column: column_index,
+        })
+    }
+}
```
jyn514: This would need to have a `number()` method to support #157.

codespan-reporting/src/term/views/locus.rs
```diff
@@ -13,31 +12,30 @@ use crate::term::Config;
...
         write!(
             writer,
-            "{file}:{line}:{column}",
-            file = PathBuf::from(self.file_name).display(),
-            line = self.location.line.number(),
-            column = self.location.column.number(),
+            "{origin}:{line}:{column}",
+            origin = self.origin,
+            line = self.location.line + 1,
+            column = self.location.column + 1,
```
jyn514: This would need to use the proposed `number()` API to support #157.

codespan-reporting/src/diagnostic.rs
jyn514: Whoops, misread the API - this is not user facing, it just has the `(line_index, column_index)` pair.

brendanzab:
So yeah, this doesn't actually fix #157, but I think I'll merge it nonetheless, as it's still helpful!

jyn514:
> So yeah, this doesn't actually fix #157, but I think I'll merge it nonetheless, as it's still helpful!
 
My concern wasn't that it doesn't fix #157, but rather that it doesn't _allow_ fixing #157, because there's no way to ask the `Line` what line it belongs to (regardless of whether that's `index + 1` or something more complicated). I'm concerned that this will make it harder to add the functionality in the future.

brendanzab:
Ahh, sorry, wouldn't have merged if you were that worried! I'm definitely happy if we alter the `Files` API, or the `Line` struct before we publish 0.9 in a way that would support it.

PR #168, opened 2020-02-26
Don’t highlight underlined source with mark color

brendanzab:
This should make it easier to implement grouped labels in the future (see #100), as we won’t have to deal with overlapping colours in the source - only overlapping underlines!

### DIFF
--- a/codespan-reporting/src/term/views/source_snippet.rs
+++ b/codespan-reporting/src/term/views/source_snippet.rs
@@ -133,7 +133,8 @@ impl<'a, F: Files> SourceSnippet<'a, F> {
             let start_line = line(start_line_index).expect("start_line");
             let end_line = line(end_line_index).expect("end_line");
 
-            let label_style = mark_style.label_style(config);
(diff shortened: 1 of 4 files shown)


jyn514:
Can you give a screenshot of what this looks like before and after?

brendanzab:
Sure!

Before:
Screen Shot 2020-02-26 at 4 20 27 pm

After:
Screen Shot 2020-02-26 at 4 20 03 pm

brendanzab:
A more involved example:

Before:
Screen Shot 2020-02-26 at 4 25 26 pm

After:
Screen Shot 2020-02-26 at 4 25 05 pm

jyn514:
Hmm, I don't have a strong preference either way. I do think the case looks better without highlighting though. If this makes #100 easier I say go for it :)

brendanzab:
Yeah, I could always make it so that the primary highlight takes precedence, but yeah, I'd rather just simplify the code more \in an attempt to make my life easier 😅

PR #169, opened 2020-02-26
Add SimpleFile and SimpleFiles structs

brendanzab:
This adds the `SimpleFile` and `SimpleFiles` structures to the `codespan_reporting` crate to help new users get started with their programming languages. It also inverts the dependency between `codespan` and `codespan_reporting`. That way users of `codespan_reporting` no longer have to depend on `codespan` if they don’t want to.


codespan-reporting/src/lib.rs
```diff
@@ -1,4 +1,285 @@
...
+    line_starts: Vec<usize>,
+}
+
+fn line_starts<'a>(source: &'a str) -> impl 'a + Iterator<Item = usize> {
+    std::iter::once(0).chain(source.match_indices('\n').map(|(i, _)| i + 1))
+}
+
+impl<Origin, Source> SimpleFile<Origin, Source>
+where
+    Origin: std::fmt::Display,
```
jyn514: I like this API. Currently I store both the absolute path in the `Source` struct and the filename in Files, this would allow me to store both in Origin instead which semantically makes more sense :)

codespan-reporting/src/lib.rs
```diff
@@ -1,4 +1,285 @@
...
+        })
+    }
+}
+
+/// A single source file.
+///
+/// This is useful for simple language tests, but it might be worth creating a
+/// custom implementation when a language scales beyond a certain size.
+#[derive(Debug, Clone)]
+pub struct SimpleFile<Origin, Source> {
```
jyn514: Do you plan to add more implementations of `Files` in the future? If so it might be worth making this a submodule so `lib.rs` doesn't get cluttered.

codespan/Cargo.toml
```diff
@@ -11,10 +11,11 @@ documentation = "https://docs.rs/codespan"
 edition = "2018"
 
 [dependencies]
+codespan-reporting = { path = "../codespan-reporting", version = "0.8.0" } # CODESPAN
```
jyn514: Hmm ... Is there a way to make this conditional on a feature, so that I don't have to depend on codespan-reporting unless I use it? A default-on feature would be fine.

brendanzab: Yes, we could do this!

brendanzab: Done in 77160d0 :)

codespan-reporting/src/lib.rs
brendanzab: Hmm I could do that!

brendanzab: I was thinking of re-exporting the diagnostic and files stuff from the top level FWIW, but I might do that in a separate PR.

brendanzab:
Not sure if you'd like to have a look at this too @Marwes?

PR #172, opened 2020-02-29
Add useful impls 

Lapz:


codespan-reporting/src/diagnostic.rs
```diff
@@ -80,7 +80,7 @@ impl<FileId> Label<FileId> {
 
 /// Represents a diagnostic message that can provide information like errors and
 /// warnings to the user.
-#[derive(Clone)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
```
jyn514: This seems a little large to derive Debug on ... can I ask your use case?

codespan/src/file.rs
```diff
@@ -330,7 +330,7 @@ where
 }
 
 /// A file that is stored in the database.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
```
jyn514: Does it make sense to compare files like this? `Eq` will be expensive and I'm not sure how useful it will be.

codespan-reporting/src/diagnostic.rs
```diff
@@ -16,7 +16,7 @@ use std::ops::Range;
 /// assert!(Severity::Warning > Severity::Note);
 /// assert!(Severity::Note > Severity::Help);
 /// ```
-#[derive(Copy, Clone, PartialEq, Hash, Debug)]
+#[derive(Copy, Clone, PartialEq, Hash, Debug, Eq)]
```
jyn514: Good call, thanks :)

codespan-reporting/src/diagnostic.rs
Lapz: The debug one I can actually remove. My main use case for the impls is salsa. Salsa requires that arguments for a query must implement hash, eq and partialEq

jyn514: This seems expensive to be in a salsa query, wouldn't it be faster to just recompute it instead of comparing all the strings to one another?

codespan/src/file.rs
brendanzab: Yeah, I'm not sure about having these impls. 🤔

codespan-reporting/src/diagnostic.rs
brendanzab: Thanks, this is a handy one! Could also maybe do with `Ord`?

Lapz:
Yeah I've realised that some the impls could be removed and that I should
have improved my code instead.

brendanzab:
No worries! I see you're also interested in Salsa! Note that I have been working on making it easier to implement custom `Files` databases in my most recently merged PRs - this will let you implement your own based on the `codespan_reporting::files::Files` trait. It's not yet up on crates.io though, but maybe it might be of use to you!

Lapz:
Kl I'll hold off with this then

PR #176, opened 2020-03-05
Begin decoupling rendering from layout

brendanzab:
We've not yet fully decoupled rendering from layout (which should make it easier to implement #100), but this gets us a little closer. It also greatly simplifies the rendering code, which hopefully makes it a bit easier to understand!

Before:
Before Screen Shot

After:
After Screen Shot


brendanzab:
Closing in favour of #183

PR #179, opened 2020-03-07
Make Diagnostic Eq + Hash

tmandry:
In order to use codespan with salsa, Diagnostics have to implement Eq and Hash. I don't see any downsides in adding these derives.

### DIFF
--- a/codespan-reporting/src/diagnostic.rs
+++ b/codespan-reporting/src/diagnostic.rs
@@ -16,7 +16,7 @@ use std::ops::Range;
 /// assert!(Severity::Warning > Severity::Note);
 /// assert!(Severity::Note > Severity::Help);
 /// ```
-#[derive(Copy, Clone, PartialEq, Hash, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub enum Severity {
     /// An unexpected bug.
@@ -50,7 +50,7 @@ impl PartialOrd for Severity {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub enum LabelStyle {
     /// Labels that describe the primary cause of a diagnostic.
@@ -60,7 +60,7 @@ pub enum LabelStyle {
 }
 
 /// A label describing an underlined region of code associated with a diagnostic.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub struct Label<FileId> {
     /// The style of the label.
@@ -108,7 +108,7 @@ impl<FileId> Label<FileId> {
 
 /// Represents a diagnostic message that can provide information like errors and
 /// warnings to the user.
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub struct Diagnostic<FileId> {
     /// The overall severity of the diagnostic


brendanzab:
Just curious, why would you need to have Salsa manage `Diagnostic`s? It could be quite expensive to compute the hashes and equalities of these, which seems to go against what you'd want from Salsa? 🤔

brendanzab:
Here's a similar PR: #172 - I'm wondering if we need to provide guidance in our docs for this kind of use case 🤔

jyn514:
I'm ok with `impl Eq for Severity` but I think the rest aren't very useful for the reasons discussed in the other PR.

brendanzab:
It might be that we need an example showing a more involved example using Salsa in the examples. I also want to encourage users to make language-specific reporting message types, then convert them to diagnostics only if they need to print to the console, but again, I think that would require documentation improvements!

tmandry:
It's certainly true that doing all these comparisons is going to be costly. I was just looking for a way to get up and running with some kind of incremental computation. Maybe it's not a good idea to encourage this.

For using salsa with diagnostics, there are a few options I see. All of these assume we want to "keep going" even if there are errors.

1. Make a type for every kind of error, and generate the strings needed at the time of reporting. Really nice and clean because it allows you to use diagnostics as lightweight values, but IMO has too much boilerplate to add a new kind of diagnostic. I think rust-analyzer does this.
1. Build up "real" diagnostics inline with the query implementation, and find a way to collect these from dependent queries while avoiding deep copies and comparisons. Build up tree-like structures and use tricks like `Arc::ptr_eq`, maybe.

Is there anything else I'm not seeing?

Based on the comment above, I suspect you want to encourage the first approach. Do you have any ideas on how to reduce the boilerplate required? Format strings embedded in the code are really hard to beat from that standpoint.

tmandry:
My current preference is (2) above, but I am having a hard time coming up with a way of doing it intelligently. Ideally, the properties I want are:

- Any query collects the diagnostics of dependent queries, plus its own diagnostics, in some kind of tree-like structure
- Diagnostics cannot be inspected, so the result of a query does not actually depend on them (it can know if there are any fatal errors, but that can always be encoded in the "value")
- A query close to the input which changes only its diagnostics results in no "heavyweight" downstream queries being recomputed (recomputing the diagnostics "tree" is fine)

Sadly, I can't think of a way to do this with salsa that isn't error-prone.

brendanzab:
Thanks for your input, I really appreciate it! I guess looking forward I'm guessing that users may eventually have to deal with things like:

- localisation
- salsa queries
- JSON backends for screen readers
- different back-ends eg:
    - the LSP tends to demand a simpler style of diagnostic that is less compatible with the rich, rustc-style of error that codespan support
    - structured editors might allow for richer errors than the LSP or `codespan-reporting`

Ideally we might want to support a user experience that allows people to gradually work up to those concerns without overwhelming them all at once. I'm still pretty wary of encouraging people to use inline format strings however, but I'm willing to be convinced, even if it's not something I do personally!

Should we make an issue for this?

brendanzab:
I'm wondering if there's a workaround that's available to you in the short term. I don't want to block your work. 😅

brendanzab:
One thing that could help with this PR is to have a small Salsa example in the `codespan-reporting/examples` folder. This might help to evaluate the verbosity of different approaches, and would be helpful for me, seeing as I've not yet delved deep into salsa and have trouble empathising with some of the pain points!

tmandry:
Interesting, yeah those are all important concerns. I'm coming from working mostly in rustc and it certainly hasn't solved all of them yet :) 

localization: I have seen it done with inline strings in frameworks like Qt, which has a `tr` macro and has a tool which collects all uses of it from your source code.

JSON/screen reader: seems possible to support with the current structure of `Diagnostic`

adapting to less-rich backends: not sure, maybe there's a general way of "lowering" a richer `Diagnostic` struct to something simpler, but you might want to be able to customize this behavior

It might be possible to write some kind of macro which looks like a (fancy) format string, but does all kinds of logic underneath.

The salsa concerns (for the second approach) have more to do with salsa itself than with codespan, I think. I [just asked](https://salsa.zulipchat.com/#narrow/stream/145099-general/topic/.22monadic.22.20computation) in the salsa zulip if this pattern is supported, but my thinking is that it will require changes to salsa.

brendanzab:
Yeah, for localisation I'm thinking something like Druid's `LocalizedString` (added in xi-editor/druid#124) could be handy. But I'm pretty sure I'd just want that to be shown in an example, rather than hard-coding it as a dependency in `codespan`.

Thanks for following up too! I'm excited to see if we can push this support even further - sorry if I'm being a bit of a stick in the mud here. Salsa is an important use case, and has been a big driver of my recent changes to codespan! ❤️

tmandry:
> Thanks for following up too! I'm excited to see if we can push this support even further - sorry if I'm being a bit of a stick in the mud here. Salsa is an important use case, and has been a big driver of my recent changes to codespan! ❤️

Not at all, I really appreciate that you've been working to make the API more flexible. I'm definitely an early adopter when it comes to using codespan with salsa and we're still figuring it out.

I don't really want to merge these changes at this point, just want to find the best way forward 😄

tmandry:
I think it might be possible to model something like (2) in salsa, although it is a bit hacky and requires two queries for each bit of logic (macros might help). I posted details in that zulip thread.

> Yeah, for localisation I'm thinking something like Druid's `LocalizedString` (added in [xi-editor/druid#124](https://github.com/xi-editor/druid/pull/124)) could be handy. But I'm pretty sure I'd just want that to be shown in an example, rather than hard-coding it as a dependency in `codespan`.

That does look promising. You could make Diagnostic generic on the string type too, perhaps? Although in my case, I'd probably do the localization step as I build the string. I don't see a whole lot of merit in lazy computation when basically every diagnostic that gets built will get displayed, but that may not be the case for every application.

brendanzab:
Gonna close this for now.

PR #188, opened 2020-03-09
Make codespan-reporting more prominent in README

brendanzab:
Since we are trying to now guide people to use `codespan-reporting` directly without relying on `codespan`, it makes sense to make it more prominent in the README.


brendanzab:
I'm not yet sure if I want to go so far as to mark `codespan` and `codespan-lsp` as `#![deprecated]` - we might want to have a longer consultation period with the community before that happens. I've just reached the point where I don't trust that we can adequately support a file database in a way that is general across all possible use cases.

brendanzab:
Ooooh, so happy with [the SVG preview](https://github.com/brendanzab/codespan/blob/4474072ae6b87b2be9cd248c47ab5587cccf08d6/codespan-reporting/assets/sample.svg)! 😍

PR #195, opened 2020-03-10
Add an example of creating a custom file backends

brendanzab:
There has been some confusion over how hard/how to make custom implementations of the `Files` trait. I think it's probably good if we provide an example of what this looks like.

### DIFF
--- a/codespan-reporting/examples/custom_files.rs
+++ b/codespan-reporting/examples/custom_files.rs
@@ -1,0 +1,178 @@
+//! This example shows how to implement a simple custom file database for use
+//! with `codespan_reporting`.
+//!
+//! This implements a custom database that uses 32-bit file-ids.
+
+use codespan_reporting::diagnostic::{Diagnostic, Label};
+use codespan_reporting::term;
+use codespan_reporting::term::termcolor::{ColorChoice, StandardStream};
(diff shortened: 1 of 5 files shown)


codespan-reporting/examples/custom_files.rs
```diff
@@ -0,0 +1,104 @@
...
+    todo!()
+}
+
+/// A file that is backed by an `Arc<String>`.
+#[derive(Debug, Clone)]
+struct ArcFile {
+    /// The name of the file.
+    name: String,
+    /// The source code of the file.
+    source: Arc<str>,
```
brendanzab: Perhaps this is a bit much of a jump in complexity? Perhaps we should just use a `String` here?

codespan-reporting/examples/custom_files.rs
```diff
@@ -0,0 +1,104 @@
...
+//!
+//! This implements a custom database that uses `Arc<str>`s for source storage,
+//! and 32-bit file-ids.
+
+use codespan_reporting::files;
+use std::ops::Range;
+use std::sync::Arc;
+
+fn main() {
+    todo!()
```
brendanzab: Still need to add an example here.

Issue #102, opened 2019-06-24
Add examples of working with other libraries


brendanzab:
It would be nice to show how codespan can be used in conjunction with other libraries. Some ones we might want to show are:

- [ ] [LALRPOP](https://github.com/lalrpop/lalrpop/)
- [ ] [combine](https://github.com/Marwes/combine)
- [ ] [Pest](https://pest.rs/)
- [ ] [Nom](https://github.com/Geal/nom)

Stupremee:
I'm currently working on a [project](https://github.com/Stupremee/kaleidoscope.rs) that uses `codespan-reporting` and `salsa`.

brendanzab:
Oh wow, this looks like a nice thing to learn from in general!

Kixiron:
I made a [minimal example](https://github.com/Kixiron/codespan-salsa) of using `salsa` & `codespan-reporting` to get on-demand computation (so you only calculate spans and stuff for the errors you actually emit)

PR #196, opened 2020-03-10
Add peg example

brendanzab:
As part of #102, this adds an example using the peg parser generator crate.

Example session:

Example session


brendanzab:
Blegh, `rustyline` requires Rust 0.40 [due to `#[non_exhaustive]`](https://caniuse.rs/features/non_exhaustive).

Issue #180, opened 2020-03-08
Release new version of codespan-lsp


ebkalderon:
There are some critical fixes and corrections available in `lsp-types` 0.73.0, which was recently released to Crates.io. However, the latest stable release of `codespan-lsp`, which is version 0.8.0 at the time of writing, currently only supports ^0.70. It would be awesome if `codespan-lsp` could be bumped to accommodate this new release.

Upgrading to the latest version of `lsp-types` would include the following fixes:

* https://github.com/gluon-lang/lsp-types/pull/139
* https://github.com/gluon-lang/lsp-types/pull/141
* https://github.com/gluon-lang/lsp-types/pull/143
* https://github.com/gluon-lang/lsp-types/pull/145
* https://github.com/gluon-lang/lsp-types/pull/146
* https://github.com/gluon-lang/lsp-types/pull/148
* https://github.com/gluon-lang/lsp-types/pull/149

Some third-party projects like `tower-lsp` depend directly on `lsp-types` alone, but are mindful to avoid upgrading to the latest release in case downstream users are relying on `codespan-lsp` in their own projects. If a new version of `codespan-lsp` were released to Crates.io with support for `lsp-types` ^0.73, `tower-lsp` would be able to safely update its dependency on `lsp-types` as well.

Additionally, downstream issue https://github.com/ebkalderon/tower-lsp/issues/10 is currently blocked on `lsp-types` pull requests https://github.com/gluon-lang/lsp-types/pull/143, https://github.com/gluon-lang/lsp-types/pull/148, and https://github.com/gluon-lang/lsp-types/pull/149, which were all merged post-0.70.2.

I'm sorry to be bugging you for new releases relatively often, @brendanzab. You're doing great work on `codespan`, and I see you've landed some great improvements to `codespan` and `codespan-reporting` in `master` already, so I apologize in advance if this request complicates your own release schedule at all. This task isn't urgent; feel free to take your time and release as you see fit!

brendanzab:
Thanks for letting me know and the kind issue! I'll want to release a new version coinciding with some of the fixes I've got in flight in #176. This should be done in the next few days! As noted at https://github.com/gluon-lang/lsp-types/issues/117#issuecomment-596278627 it's a huge burden to try and keep `codespan-lsp` and `codespan-reporting` versioned in lockstep when `codespan-lsp` needs frequent updates. I might need to reconsider my strategy on this in the long term. 🤔

brendanzab:
[...]

PR #201, opened 2020-03-11
Start using a version range for lsp-types

brendanzab:
Based on discussions on #180 and gluon-lang/lsp-types#117, this switches us to use a version range (`>=0.70,<0.74`) for our dependency on `lsp-types`, seeing as we only use a couple of types from this repository (`lsp_types::{Position, Range}`) that don’t change very often. This will mean that we can make releases of `codespan-lsp` that update the maximum version _without_ breaking backwards compatibility for consumers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,17 @@ jobs:
         with:
           command: check
           args: --manifest-path "codespan-lsp/Cargo.toml"
+      - name: Switch to minimal `lsp-types` version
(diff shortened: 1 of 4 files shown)


Marwes:
It would be useful to run a check build in travis with the min and max versions for the range. Either with the unstable min versions flag or by setting it explicitly `cargo update --precise lsp-types:0.70`

brendanzab:
Good idea! Done!

PR #203, opened 2020-03-12
Avoid rendering multiple source lines when labels are on one line

brendanzab:
As another step towards #100, this means that this:
```
   ┌── one_line.rs:3:5 ───
   │
 3 │     v.push(v.pop().unwrap());
   │     - first borrow later used by call
   ·
 3 │     v.push(v.pop().unwrap());
   │       ---- first mutable borrow occurs here
   ·
 3 │     v.push(v.pop().unwrap());
   │            ^ second mutable borrow occurs here
   │
```
…is now rendered as:
```
   ┌── one_line.rs:3:5 ───
   │
 3 │     v.push(v.pop().unwrap());
   │     - first borrow later used by call
   │       ---- first mutable borrow occurs here
   │            ^ second mutable borrow occurs here
   │
```
We’ll eventually want to use vertical lines like rustc does to make this clearer, eg:
```
  --> $DIR/one_line.rs:3:12
   |
LL |     v.push(v.pop().unwrap());
   |     - ---- ^ second mutable borrow occurs here
   |     | |
   |     | first borrow later used by call
   |     first mutable borrow occurs here
```

### DIFF
--- a/codespan-reporting/src/term/views.rs
+++ b/codespan-reporting/src/term/views.rs
@@ -41,15 +41,15 @@ where
 
         // Group marks by file
         for label in &self.diagnostic.labels {
-            let start_index = files.line_index(label.file_id, label.range.start).unwrap();
-            let end_index = files.line_index(label.file_id, label.range.end).unwrap();
-            let end_number = files.line_number(label.file_id, end_index).unwrap();
+            let start_line_index = files.line_index(label.file_id, label.range.start).unwrap();
+            let end_line_index = files.line_index(label.file_id, label.range.end).unwrap();
+            let end_line_number = files.line_number(label.file_id, end_line_index).unwrap();
 
             // The label spans over multiple lines if if the line indices of the
             // start and end differ.
-            let is_multiline = start_index != end_index;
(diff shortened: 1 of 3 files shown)


brendanzab:
Any thoughts on whether I should merge this? Bar 🚀 ?

Marwes:
Something, something "we are in the future now", I think :) .

While it isn't perfect I would say it is not really worse than the current way of rendering either even. So if it brings us closer to something like rustc's rendering it may still be worth merging as an incremental step.

brendanzab:
Awesome, and thanks for keeping on top with some of these changes!

I guess one of my concerns is that is _does_ muddy up the rendering code a little, which could lead to future changes muddying that up even more… which could result in an unmaintainable mess. I guess it's more something to watch though.

I do have a hunch that I might want another step before actually calling the rendering methods. Like, to group together 'islands' of labels that are in the same bit of source code. But yeah, still not sure. This stuff is hard for my lil brain to figure out. 😖

brendanzab:
I've decided to have a think about this before merging. Don't want the rendering code to get _too_ unmaintainable.

PR #207, opened 2020-03-19
Implement some conversions for indexes and spans

brendanzab:
As I've said in the past, I'd like to continue to nudge people towards using `codespan-reporting` with `codespan_reporting::files::SimpleFiles` or their own implementation of `codespan_reporting::files::Files`, but this could help people who are currently using `codespan` more work with the new diagnostic data types in a cleaner way.

### DIFF
--- a/codespan/src/span.rs
+++ b/codespan/src/span.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::ops::Range;
 
-use crate::ByteIndex;
+use crate::{ByteIndex, RawIndex};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serialization", derive(Deserialize, Serialize))]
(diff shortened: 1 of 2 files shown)


codespan/src/span.rs
```diff
@@ -137,6 +137,15 @@ where
     }
 }
 
+impl<I> From<Span> for Range<I>
```
brendanzab: Dang - this seems valid on nightly at least! 🤔

jyn514: The error makes sense to me? `impl From Span for Range<I>` means that it can be turned into a range over _any_ type, including types not in std. Try `impl From Span for Range<usize>` instead.

brendanzab: I dunno, I think it should work, but the rule might have been loosened recently 🤔

brendanzab: I'll just do `Range<usize>` though

ebkalderon:
Was literally just looking for a `Range<usize>` conversion so I could pass a `codespan::Span` to `codespan_reporting::diagnostic::Label::primary()`. Thanks for adding this, @brendanzab!

brendanzab:
Just published this in v0.9.1!

Issue #156, opened 2020-02-01
Take T: Display instead of String for Label and Diagnostic


jyn514:
Currently, I'm writing code that looks like this:

```rust
     for warning in warnings {
        let label = Label::new(file, warning.location.span, warning.data.to_string());
        Diagnostic::new_warning(warning.data.to_string(), label);
        // TODO: emit warning
    }
```

My warnings are fairly compact error enums. If `new_warning` and `Label::new` took `Display` instead, they could
- take up less space in memory, and
- I wouldn't have to write `to_string()` so much

PR #217, opened 2020-03-27
Change `.with_message()` method to take more generic message type

oOBoomberOo:
Implement #156 
This only affect the `.with_message()` method, the underlying data is still String.


jyn514:
Would it be possible to make the struct generic as well so you don't have to store the full string in memory the whole time? If so I could generate diagnostics at the spot they're issued instead of just before they're displayed. I have several errors that are just a unit variant as a struct but fairly long if you show the error message.

oOBoomberOo:
Look like these are the only place that needs `message` value ([#1](https://github.com/brendanzab/codespan/blob/master/codespan-reporting/src/term/views.rs#L247), [#2](https://github.com/brendanzab/codespan/blob/master/codespan-reporting/src/term/views.rs#L247), [#3](https://github.com/brendanzab/codespan/blob/master/codespan-reporting/src/term/views.rs#L247)) and it shouldn't be much of a trouble to change them.
The only problem would be that you will be exposing more generics which could sometime be annoying to write.

brendanzab:
I guess a tad concerned about the performance implications of this, and how easy it would be to back away from later. Not sure how important it is though. 🤔

brendanzab:
I'm definitely interested in if we can reduce intermediate allocations too. One complicating issue is that it might be important in some places to measure the lengths of messages, or distribute them over lines, or maybe do layout, eg. using the `pretty` crate.

brendanzab:
Sorry, I forgot to comment on the issue when it was made!

Bad maintainer, bad! 😅

brendanzab:
I'm willing to be convinced too. Interested if @Marwes has any thoughts

brendanzab:
(woops, that close was a mistake)

oOBoomberOo:
Maybe we could encourage the use of error enum like in `reusable_diagnostic` example? You can control when to generate the diagnostic (and string).

brendanzab:
I wonder if we could take a `&dyn Display` we'd take a vtable hit, but not a dynamic allocation hit. And taking references would encourage folks not to keep the `Diagnostic` hanging around for long? Just a thought.

oOBoomberOo:
I don't think that will work since that will require the `Display` struct to live until `Diagnostic` is reported.
Considered this example:
```rust
enum Error {
    Foo(PathBuf)
}

impl Error {
    fn report(&self) -> Diagnostic<()> {
        match self {
           Error::Foo(path) => {
               let path = path.display();
               Diagnostic::error().with_message(path)
           }
        }
    }
}
```
`path` is generated locally here and would be immediately drop after that while `Diagnostic` will be use in the outer scope.

brendanzab:
Ahh dang - I guess in that case it would have to be a `Cow`… bleh 🤔

jyn514:
FWIW if the idea is for the diagnostic to be generated as late as possible I don't think we need to generalize this too much, we can just leave it as taking String.

brendanzab:
Gonna close this for now - feel free to make an issue to discuss this further!

Johann150:
I think we should merge this, maybe change from `Display` to `ToString` (anything that implements `Display` already automatically implements `ToString` <https://doc.rust-lang.org/std/string/trait.ToString.html#impl-ToString-4>, but not the other way around).

Deferring the generation of the error message as long as possible is too complex though IMHO.

Johann150:
While we're at it, would it be a good idea to apply the same changes to `Diagnostic::with_code`?

Issue #224, opened 2020-04-01
Panics when sending invalid byte indexes to codespan-reporting.


ratmice:
In issue #189 I ran into an issue with a bug in a lexer, where it was giving byte indexes not on a char boundary, passing those on to codespan-reporting resulted in panics, It was proposed there to open a separate issue for that.

I have run into another instance of this, which (I haven't fully looked into yet).  With quite a different cause.  I uploaded a repostory here containing a [json-pop](https://github.com/ratmice/json-pop), 

In this new case, we're trying to actually display a unicode error that the input contains invalid utf8, so when trying to display the source of the error then causes the error display to fail.  

I don't have a very good testcase for it yet as the testsuite isn't running through codespan-reporting yet. You'll find the string [here](https://github.com/ratmice/json-pop/blob/41f8d4d7228401b083b7aace5ca7e1a8c132e223/src/lib.rs#L80), and it fails with the non-codespan println! error reporting the testsuite currently goes through.

jyn514:
How can a `String` be invalid UTF8?

ratmice:
it's an `&str`, but good question none the less.  Perhaps its getting through from an `from_utf8_unchecked` call in a dependency being the start of a valid sequence? :confused:

jyn514:
I agree the panics aren't ideal, but what should codespan do instead? It's not possible to print only part of a character.

ratmice:
It's a good question, while admittedly not yet being very familiar with codespan internals,
if it would be possible to get a slice via the `buffer[range.start..]` and then iterate over `char_indices()` until it finds a character index `>= range.end` and use that for displaying the character?

PR #225, opened 2020-04-01
Add padding to the source range if it doesn't end on a char boundary.

ratmice:
I'm not sure if this is the best, but this avoids a panic when given a bad location range ending in the middle of a unicode character.

This fixes all the panics I see in #224 when running [json-pop](https://github.com/ratmice/json-pop) through the invalid cases at [JSONTestSuite](https://github.com/nst/JSONTestSuite/tree/master/test_parsing)

### DIFF
--- a/codespan-reporting/src/term/renderer.rs
+++ b/codespan-reporting/src/term/renderer.rs
@@ -469,12 +469,14 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
         range: Range<usize>,
         message: &str,
     ) -> io::Result<()> {
-        let space_len = self.config.width(&source[..range.start]);
+        let space_source = slice_at_char_boundaries(source, 0..range.start);
+        let space_len = self.config.width(space_source);
         write!(self, " {space: >width$}", space = "", width = space_len)?;
         self.set_color(self.styles().label(severity, label_style))?;
+        let source = slice_at_char_boundaries(source, range);
         // We use `usize::max` here to ensure that we print at least one
         // label character - even when we have a zero-length span.
-        for _ in 0..usize::max(self.config.width(&source[range.clone()]), 1) {
+        for _ in 0..usize::max(self.config.width(source), 1) {
             write!(self, "{}", self.chars().single_caret_char(label_style))?;
         }
         if !message.is_empty() {
@@ -696,5 +698,86 @@ impl<'writer, 'config> WriteColor for Renderer<'writer, 'config> {
 
     fn is_synchronous(&self) -> bool {
         self.writer.is_synchronous()
+    }
+}
+
+/// Searches for character boundary from byte_index towards the end of the string.
+fn closest_char_boundary(s: &str, byte_index: usize) -> usize {
+    let length = s.len();
+    for index in byte_index..=length {
+        if s.is_char_boundary(index) {
+            return index;
+        }
+    }
+    length
+}
+
+/// Searches for character boundary from byte_index towards the start of the string.
+fn closest_char_boundary_rev(s: &str, byte_index: usize) -> usize {
+    for index in (0..=byte_index).rev() {
+        if s.is_char_boundary(index) {
+            return index;
+        }
+    }
+    0
+}
+
+/// Finds a valid unicode boundaries looking from `range.start` towards the beginning of the string.
+/// From `range.end` towards the end of the string. Returning a `&str` of all characters
+/// that overlapping the range.
+fn slice_at_char_boundaries<'a>(s: &'a str, range: Range<usize>) -> &'a str {
+    let start = closest_char_boundary_rev(s, range.start);
+    let end = closest_char_boundary(s, range.end);
+    &s[start..end]
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::iter::repeat;
+
+    #[test]
+    fn test_boundary() {
+        let s = "🌞";
+        assert_eq!(closest_char_boundary(s, 0), 0);
+        assert_eq!(closest_char_boundary_rev(s, 0), 0);
+        for i in 1..s.len() {
+            assert_eq!(closest_char_boundary_rev(s, i), 0);
+            assert_eq!(closest_char_boundary(s, i), s.len());
+        }
+    }
+
+    #[test]
+    fn test_boundaries() {
+        let s = "🌑🌒🌓🌔";
+        let individually = ["🌑", "🌒", "🌓", "🌔"];
+
+        let mut expect = Vec::new();
+        // [(0, 0, "", ""),
+        //  (0, 4, "🌑", "🌑"), repeated 4 times,
+        //  (4, 4, "", ""), once
+        //  (4, 8, "🌒", "🌒"), repeated 4 times, and so on (+4, +4)
+        //  ...
+        //  (16, 16, "", ""); 21]
+        expect.push((0, 0, "", ""));
+        for (idx, &char) in individually.iter().enumerate() {
+            let n = char.len();
+            assert_eq!(n, 4);
+            let expected_start = (idx % n) * n;
+            let expected_end = (idx % n) * n + n;
+            expect.extend(repeat((expected_start, expected_end, char, char)).take(n - 1));
+            expect.push((expected_end, expected_end, "", ""));
+        }
+
+        // drop mut.
+        let expect = expect;
+        let mut found = Vec::new();
+        for i in 0..=s.len() {
+            let sliced = slice_at_char_boundaries(s, i..i);
+            let prev = closest_char_boundary_rev(s, i);
+            let next = closest_char_boundary(s, i);
+            found.push((prev, next, &s[prev..next], sliced));
+        }
+        assert_eq!(found, expect);
     }
 }
--- a/codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
+++ b/codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
@@ -1,0 +1,16 @@
+---
+source: codespan-reporting/tests/term.rs
+expression: TEST_DATA.emit_no_color(&config)
+---
+error[E01]: cow may not jump during new moon.
+
+   ┌─ moon_jump.rs:1:1
+   │
+ 1 │ 🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄
+   │   ^^ Invalid jump
+   │   -- Cow range does not start at boundary.
+   │   ------ Cow does not start or end at boundary.
+   │     -- Cow range does not end at boundary.
+   │
+
+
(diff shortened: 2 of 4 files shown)


ratmice:
Well, it isn't very pretty, but this does make it through `$ for i in ../../JSONTestSuite/test_parsing/n_*; do echo $i && cargo run --example json-pop --features pretty_errors -- --color never <$i; done 2>&1 | grep panic` without errors... Too terrible?

codespan-reporting/src/term/renderer.rs
```diff
@@ -472,9 +472,34 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+            let mut pos = range.end + 1;
+            while pos < total_len {
+                if source.is_char_boundary(pos) {
+                    return range.start..pos;
+                }
+                pos += 1;
+            }
+
+            return range.start..pos;
+        }
```
brendanzab, line 495: I wonder if we could extract this into a private, local function, and give it some tests? I'm thinking something like:

```rust
fn closest_char_boundary(s: &'a str, byte_index: usize) -> usize {
    ⋮
}

fn slice_at_char_boundaries<'a>(s: &'a str, range: Range<usize>) -> &'a str {
    let start = closest_char_boundary(s, range.start);
    let end = closest_char_boundary(s, range.end);
    &s[start..end]
}
```

I had similar issues with trying to make `files::column_index` look nice, but eventually ended up with something that I thought was pretty good:

https://github.com/brendanzab/codespan/blob/44a88ee938cfbdff0f7ea84b86b256b6aa5c4aab/codespan-reporting/src/files.rs#L126-L156

Perhaps it could be worth looking at it for some ideas on how to clean this up? If not, it's not a big deal, I can always have a look later.

brendanzab:
Sorry for the late reply! This looks mighty useful! It would be really cool to have an integration test for this!

ratmice:
No worries, :hourglass_flowing_sand: , I will have a look at adding tests and see if I can't clean it up some.

codespan-reporting/src/term/renderer.rs
ratmice: I moved them to a file utils.rs, noticing that there are a couple of direct slicings in e.g. view.rs.

This is going to need another round, I came up with a couple cleaner solutions,
one is combinator-esque and the one I felt was a little clearer just uses a for loop.
I left the combinator one commented out in case that is preferred.

ratmice:
> Sorry for the late reply! This looks mighty useful! It would be really cool to have an integration test for this!

So, I added some tests, they even caused a panic which my parser didn't exercise (related to start).
However, the output doesn't seem quite right, so I hadn't added the .snap file.

I didn't because i'm not quite sure what the correct output will look like in regards to the ordering of lines eventually.

codespan-reporting/src/term.rs
```diff
@@ -9,6 +9,7 @@ use crate::files::Files;
 
 mod config;
 mod renderer;
+mod utils;
```
brendanzab, line 12: Would it be possible for these to just be local in the `renderer` module? I really try to avoid `util` modules - they end up becoming kitchen sinks hehe :)

codespan-reporting/src/term/renderer.rs
```diff
@@ -469,12 +470,17 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+            .width(slice_at_char_boundaries(source, 0..range.start));
         write!(self, " {space: >width$}", space = "", width = space_len)?;
         self.set_color(self.styles().label(severity, label_style))?;
         // We use `usize::max` here to ensure that we print at least one
         // label character - even when we have a zero-length span.
-        for _ in 0..usize::max(self.config.width(&source[range.clone()]), 1) {
+        for _ in 0..usize::max(
+            self.config.width(slice_at_char_boundaries(source, range)),
+            1,
+        ) {
```
brendanzab, line 483: Perhaps this would be nicer:
```rust
        let space_source = slice_at_char_boundaries(source, 0..range.start);
        let space_len = self.config.width(&space_source);
        write!(self, " {space: >width$}", space = "", width = space_len)?;
        self.set_color(self.styles().label(severity, label_style))?;

        let source = slice_at_char_boundaries(source, range);
        // We use `usize::max` here to ensure that we print at least one
        // label character - even when we have a zero-length span.
        for _ in 0..usize::max(self.config.width(&source), 1) {
```
Sometimes you need to work with rustfmt to get it to format nicely! :) :)

codespan-reporting/tests/term.rs
```diff
@@ -540,3 +540,41 @@ mod unicode {
     test_emit!(rich_no_color);
     test_emit!(short_no_color);
 }
+
+mod unicode_spans {
+    use super::*;
+
+    lazy_static::lazy_static! {
+        static ref TEST_DATA: TestData<'static, SimpleFile<&'static str, String>> = {
+            let moon_phases = format!("{}", r#"🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄"#);
```
brendanzab, line 549: HAHAH love these cows

brendanzab: Might need to look at cleaning up the formatting here (trailing commas and spacing) - rustfmt seems to ignore `lazy_static!` sadly :(

codespan-reporting/src/term/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+        //  (16, 16, "", ""); 18]
+        expect.push((0, 0, "", ""));
+        for (idx, &char) in individually.iter().enumerate() {
+            let n = char.len();
+            assert_eq!(n, 4);
+            let expected_start = (idx % n) * n;
+            let expected_end = (idx % n) * n + n;
+            expect.extend(repeat((expected_start, expected_end, char, char)).take(n - 1));
+            expect.push((expected_end, expected_end, "", ""));
+        }
```
brendanzab, line 89: We could potentially use property testing here, but I think this is fine for now!

codespan-reporting/src/term/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+pub(crate) fn closest_char_boundary_rev(s: &str, byte_index: usize) -> usize {
+    for index in (0..=byte_index).rev() {
+        if s.is_char_boundary(index) {
+            return index;
+        }
+    }
+    0
+}
+
+pub(crate) fn slice_at_char_boundaries<'a>(s: &'a str, range: Range<usize>) -> &'a str {
```
brendanzab, line 47: I think `pub` would suffice here seeing as the module isn't publicly exported - I try to avoid `pub(crate)` - not the biggest fan of it. 😅

codespan-reporting/tests/term.rs
```diff
@@ -540,3 +540,41 @@ mod unicode {
...
+                          Label::secondary((), invalid_start.."🐄🌑🐄".len() - 1)
+                            .with_message("Cow does not start or end at boundary.")
+
+                    ])];
+            TestData{files: file, diagnostics }
+        };
+    }
+
+    test_emit!(rich_no_color);
+    test_emit!(short_no_color);
```
brendanzab, line 579: Was this test run, and the snapshots generated with `cargo insta review`?

codespan-reporting/src/term/utils.rs
ratmice, line 47: I think this signature might eventually need to change to allow RangeTo<usize> as well in the future?

I don't see anything specific in std::convert, would that be something like Into<Range<usize>>, or just conversion by the caller?

codespan-reporting/tests/term.rs
ratmice: I think i made this match the others.

codespan-reporting/src/term/utils.rs
ratmice: I've look at property testing when I get a chance, and keep it in mind.

codespan-reporting/tests/term.rs
ratmice: well, they were generated, I hadn't gone through the insta review part (didn't know about it).
I accepted them now, but the output is sub-optimal, as the alignment is off.

I figure I could just accept them for now as it is an improvement over panicing,
but the output needs some work still I think.

codespan-reporting/src/term.rs
ratmice: I'm not so sure that the way I moved this over to render is a good way, rust really seems to want to force it into .. or term/utils/mod.rs :scream: I got it importing from `renderer` but...

codespan-reporting/src/term/utils.rs
brendanzab: I think this is fine - it's not really necessary for an internal function.

codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
```diff
@@ -0,0 +1,16 @@
...
+---
+error[E01]: cow may not jump during new moon.
+
+   ┌─ moon_jump.rs:1:1
+   │
+ 1 │ 🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄
+   │   ^^ Invalid jump
+   │   -- Cow range does not start at boundary.
+   │   ------ Cow does not start or end at boundary.
+   │     -- Cow range does not end at boundary.
```
brendanzab, line 13: I think the ordering is fine - it's not really relevant to this PR.

codespan-reporting/src/term/utils.rs
```diff
@@ -0,0 +1,77 @@
...
+    let length = s.len();
+    for index in byte_index..=length {
+        if s.is_char_boundary(index) {
+            return index;
+        }
+    }
+    length
+}
+
+pub fn closest_char_boundary_rev(s: &str, byte_index: usize) -> usize {
```
brendanzab, line 13: Why do we need a different algorithm for the end bound?

ratmice: Essentially searches for the closest boundary from the current index moving towards the left, (closest_char_boundary_rev), and the other goes from left to right for the next available boundary, both return the current character if it's a boundary, With the same algorthm for start and end I believe it would always return an empty string.

brendanzab: Wouldn't that be fine, given that we'll still render a single caret for that?

brendanzab: Hmm, I think this is fine though, was just asking! Can we put these functions and tests in the `renderer` module for now?

ratmice: It might be, I think isn't it also used for displaying the source code text, or is that a separate line range?

ratmice: sure I'll move.

ratmice: So, moved into place, while I was at it I added some doc strings.

PR #226, opened 2020-04-03
Add spacing configuration

brendanzab:
This adds three levels of spacing for rich diagnostics:

-   Cosy (default):
    ```
    error: unknown builtin: `NATRAL`

       ┌─ Data/Nat.fun:7:13
       │
     7 │ {-# BUILTIN NATRAL Nat #-}
       │             ^^^^^^ unknown builtin
       │
       = there is a builtin with a similar name: `NATURAL`

    error[E0001]: unexpected type in application of `_+_`

        ┌─ Test.fun:4:11
        │
      4 │ _ = 123 + "hello"
        │           ^^^^^^^ expected `Nat`, found `String`
        │
        ┌─ Data/Nat.fun:11:1
        │
     11 │ _+_ : Nat → Nat → Nat
        │ --------------------- based on the definition of `_+_`
        │
        = expected type `Nat`
             found type `String`

    ```

-   Compact:
    ```
    error: unknown builtin: `NATRAL`
      ┌─ Data/Nat.fun:7:13
      │
    7 │ {-# BUILTIN NATRAL Nat #-}
      │             ^^^^^^ unknown builtin
      │
      = there is a builtin with a similar name: `NATURAL`

    error[E0001]: unexpected type in application of `_+_`
       ┌─ Test.fun:4:11
       │
     4 │ _ = 123 + "hello"
       │           ^^^^^^^ expected `Nat`, found `String`
       │
       ┌─ Data/Nat.fun:11:1
       │
    11 │ _+_ : Nat → Nat → Nat
       │ --------------------- based on the definition of `_+_`
       │
       = expected type `Nat`
            found type `String`
    ```

-   Compressed:
    ```
    error: unknown builtin: `NATRAL`
      ┌─ Data/Nat.fun:7:13
    7 │ {-# BUILTIN NATRAL Nat #-}
      │             ^^^^^^ unknown builtin
      = there is a builtin with a similar name: `NATURAL`
    error[E0001]: unexpected type in application of `_+_`
       ┌─ Test.fun:4:11
     4 │ _ = 123 + "hello"
       │           ^^^^^^^ expected `Nat`, found `String`
       ┌─ Data/Nat.fun:11:1
    11 │ _+_ : Nat → Nat → Nat
       │ --------------------- based on the definition of `_+_`
       = expected type `Nat`
            found type `String`
    ```

PR #235, opened 2020-04-28
Tighten up spacing around source snippets

brendanzab:
This eliminates the spacing between the title and the locus, and the outer spacing to the left of the source numbers.

Originally we would render this:

```
error: unknown builtin: `NATRAL`

   ┌─ Data/Nat.fun:7:13
   │
 7 │ {-# BUILTIN NATRAL Nat #-}
   │             ^^^^^^ unknown builtin
   │
   = there is a builtin with a similar name: `NATURAL`

```

And now we render this:

```
error: unknown builtin: `NATRAL`
  ┌─ Data/Nat.fun:7:13
  │
7 │ {-# BUILTIN NATRAL Nat #-}
  │             ^^^^^^ unknown builtin
  │
  = there is a builtin with a similar name: `NATURAL`

```

This is an alternative to PR #226.

### DIFF
--- a/codespan-reporting/assets/readme_preview.svg
+++ b/codespan-reporting/assets/readme_preview.svg
@@ -1,4 +1,4 @@
-<svg viewBox="0 0 882 350" xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="0 0 882 335" xmlns="http://www.w3.org/2000/svg">
   <style>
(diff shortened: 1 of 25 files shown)


brendanzab:
This is what the output looks like after this change:

jyn514:
I like this a lot :) I think it does a foo

Issue #231, opened 2020-04-17
[v 0.9.2] Recent changes to label ordering make errors more confusing


tnowacki:
I find the recent change to the order in which labels displayed make a lot of error messages *really* confusing to read when displayed in the new format. 
https://github.com/brendanzab/codespan/commit/c3dd8465c762d7ead9471426d56fa539e4b324a2

For example, I find 
```
    ┌── tests/move_check/borrows/assign_local_combo_invalid.move:14:9 ───
    │
 14 │         s = S { f: 0, g: 0 };
    │         ^ Invalid assignment of local 's'
    ·
 13 │         if (cond) f = &s.f else f = &s.g;
    │                                     ---- It is still being borrowed by this reference
    │
```
A lot easier to reason through than the new version:
```
    ┌── tests/move_check/borrows/assign_local_combo_invalid.move:13:37 ───
    │
 13 │         if (cond) f = &s.f else f = &s.g;
    │                                     ---- It is still being borrowed by this reference
 14 │         s = S { f: 0, g: 0 };
    │         ^ Invalid assignment of local 's'
    │
```

Sometimes, if they are on the same line, it gets really really weird:
```
   ┌── tests/move_check/expansion/global_access_pack.move:3:13 ───
   │
 3 │         ::S { }
   │             ^ Unexpected '{'
   ·
 3 │         ::S { }
   │             - Expected '('
   │
```
Becomes
```
   ┌── tests/move_check/expansion/global_access_pack.move:3:13 ───
   │
 3 │         ::S { }
   │             - Expected '('
   │             ^ Unexpected '{'
   │
```
Do you have any suggests to how to fix this if we are on version 0.9.2?

Further context and discussion:
https://github.com/libra/libra/pull/3447

jonathandturner:
@tnowacki - I'm not sure about the multiple labels on the same line, but we talk a bit about the ordering of lines in the original design for rustc - https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html

I wouldn't say there's like a perfect solution here, but being able to see the code in the order you wrote it keeps you from having to look left multiple times to find the line numbers. It's a small time save, though I'd be curious to see if there were better ways to shave off a little time there.

brendanzab:
[...]

PR #241, opened 2020-05-14
Improve single line label rendering

### LINKED ISSUE #211: Make single-line label rendering clearer
Instead of the following (see #203):
```
   ┌── one_line.rs:3:5 ───
   │
 3 │     v.push(v.pop().unwrap());
   │     - first borrow later used by call
   │       ---- first mutable borrow occurs here
   │            ^ second mutable borrow occurs here
   │
```
...we should [render something like rustc does](https://github.com/rust-lang/rust/blob/c20d7eecbc0928b57da8fe30b2ef8528e2bdd5be/src/test/ui/codemap_tests/one_line.stderr):
```
   ┌── one_line.rs:3:5 ───
   │
 3 │     v.push(v.pop().unwrap());
   │     - ---- ^ second mutable borrow occurs here
   │     │ │
   │     │ first mutable borrow occurs here
   │     first borrow later used by call
   │
```

I'm not sure how this would interact with multiline labels though…

brendanzab:
This makes the rendering of carets smarter when there are multiple labels on the same line. Instead of this:

```
   ┌─ one_line.rs:3:5
   │
 3 │     v.push(v.pop().unwrap());
   │     - first borrow later used by call
   │       ---- first mutable borrow occurs here
   │            ^ second mutable borrow occurs here
   │
```
...we now render the following:
```
   ┌─ one_line.rs:3:5
   │
 3 │     v.push(v.pop().unwrap());
   │     - ---- ^ second mutable borrow occurs here
   │     │ │
   │     │ first mutable borrow occurs here
   │     first borrow later used by call
   │
```

Resolves #211.

### DIFF
--- a/codespan-reporting/tests/snapshots/term__same_ranges__rich_no_color.snap
+++ b/codespan-reporting/tests/snapshots/term__same_ranges__rich_no_color.snap
@@ -6,7 +6,9 @@ error: Unexpected token
   ┌─ same_range:1:5
   │
 1 │ ::S { }
-  │     ^ Unexpected '{'
-  │     - Expected '('
+  │     ^
+  │     │
+  │     Unexpected '{'
+  │     Expected '('
 
 
--- a/codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
+++ b/codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
@@ -6,13 +6,13 @@ error[E01]: cow may not jump during new moon.
   ┌─ moon_jump.rs:1:1
   │
 1 │ 🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄
-  │   ^^ Invalid jump
+  │ ^^ Invalid jump
 
 note: invalid unicode range
   ┌─ moon_jump.rs:1:1
   │
 1 │ 🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄
-  │   -- Cow range does not start at boundary.
+  │ -- Cow range does not start at boundary.
 
 note: invalid unicode range
   ┌─ moon_jump.rs:1:3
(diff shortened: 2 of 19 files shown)


brendanzab:
Current status:

Screen Shot 2020-05-17 at 4 14 52 pm

brendanzab:
Ok, this should be ready for review!

codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
```diff
@@ -6,9 +6,24 @@ error[E01]: cow may not jump during new moon.
   ┌─ moon_jump.rs:1:1
   │
 1 │ 🐄🌑🐄🌒🐄🌓🐄🌔🐄🌕🐄🌖🐄🌗🐄🌘🐄
-  │   ^^ Invalid jump
-  │   -- Cow range does not start at boundary.
-  │   ------ Cow does not start or end at boundary.
+  │ ^^ Invalid jump
```
brendanzab, line 9: These carets have changed position, due to them lying between character boundaries. If this is a big issue, we can fix this in another pass.

codespan-reporting/tests/snapshots/term__same_ranges__rich_no_color.snap
```diff
@@ -6,7 +6,9 @@ error: Unexpected token
   ┌─ same_range:1:5
   │
 1 │ ::S { }
-  │     ^ Unexpected '{'
-  │     - Expected '('
+  │     ^
+  │     │
+  │     Unexpected '{'
+  │     Expected '('
```
brendanzab, line 12: This should ultimately address the main complaint that was described in #231 by @tnowacki. cc. https://github.com/libra/libra/pull/3447

brendanzab:
Note that this still does not cover the case where single line labels overlap multiline labels. Multiline labels will still be rendered below single line labels. I still think this is a much needed improvement, however!

codespan-reporting/tests/snapshots/term__unicode_spans__rich_no_color.snap
ratmice: agree trying to deal with that in this patch would probably be too much, and there seems to be FIXME's in place.

PR #251, opened 2020-06-22
 Allow using codespan without codespan-reporting

jyn514:
This copies the implementation of `line_starts` to `codespan`. It also adds `--no-default-features` to CI.

r? @brendanzab


brendanzab:
Awesome!

PR #253, opened 2020-06-23
Restore highlighted source for primary labels

### LINKED ISSUE #206: Highlight text pointed to by labels
I noticed that codespan-reporting 0.9 is no longer highlighting the file characters pointed to by labels. For my use case I think I found that the highlighted way was preferable. Is this something that you would be open to bringing back as an option?

<br>

### codespan-reporting 0.8.0:

Screenshot from 2020-03-18 15-00-30

### codespan-reporting 0.9.0:

Screenshot from 2020-03-18 14-58-02

brendanzab:
This renders the source text marked by primary labels. 

For example:

Screen Shot 2020-06-23 at 1 32 33 pm

Notice how `num` is now coloured in red! 🎉 

Closes #206.

### DIFF
--- a/codespan-reporting/src/term/renderer.rs
+++ b/codespan-reporting/src/term/renderer.rs
@@ -30,25 +30,25 @@ pub enum MultiLabel<'diagnostic> {
     /// ```text
     /// ╭
     /// ```
-    TopLeft(LabelStyle),
+    TopLeft,
     /// Multi-line label top.
     ///
     /// ```text
     /// ╭────────────^
     /// ```
-    Top(LabelStyle, RangeTo<usize>),
+    Top(RangeTo<usize>),
     /// Left vertical labels for multi-line labels.
     ///
     /// ```text
     /// │
     /// ```
-    Left(LabelStyle),
+    Left,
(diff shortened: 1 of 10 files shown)


codespan-reporting/src/term/renderer.rs
```diff
@@ -266,12 +266,46 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+                }
+
                 match ch {
                     '\t' => (0..metrics.unicode_width).try_for_each(|_| write!(self, " "))?,
                     _ => write!(self, "{}", ch)?,
                 }
             }
+            if in_primary {
+                self.reset()?;
+            }
```
brendanzab, line 308: This whole section is rather ugly 😓 - not sure how to improve it.

Marwes: I am guessing a bit, but maybe it is better to keep a stack of styles while rendering and then just push/pop as different styles are applied.

brendanzab: Yeah, perhaps I could have some state in the renderer that handles this. Good idea! I managed to clean it up a _bit_ though.

PR #257, opened 2020-07-05
limit lines of context

### LINKED ISSUE #125: Support collapsing multi line output
Love the library, but there's an issue when using it with long multi-line spans--it shows *every* line!

I believe the lines that would need to be changed are these: https://github.com/brendanzab/codespan/blob/e2d6fc3bf504e4e9560bd898cdb5f2f740926505/codespan-reporting/src/term/views/source_snippet/mod.rs#L259-L261

image

rustc, on the other hand, will collapse multiple line spans:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0d4c7f47b06cefb15528f7106766d5f7

image

Does codespan support this?

Johann150:
This should resolve #125.

The context is limited to 3 lines after the first line or before the last line respectively. The number of lines should probably be configurable; I am still looking into that.

Does this pull request have any chance of being accepted at this time, or is this issue still on hold? @brendanzab mentioned that this should be implemented after another issue (which is now closed).


codespan-reporting/src/term/views.rs
```diff
@@ -207,8 +211,21 @@ where
...
                 for line_index in (start_line_index + 1)..end_line_index {
+                    if std::cmp::min(line_index - start_line_index, end_line_index - line_index)
+                        > renderer.context_lines()
+                        && !self.diagnostic.labels.iter().any(|label| {
+                            // check that no other label is on this line
+                            start_line_range.start <= label.range.start
+                                && label.range.start <= end_line_range.end
+                                || start_line_range.start <= label.range.end
+                                    && label.range.end <= end_line_range.end
+                        })
```
brendanzab, line 223: Perhaps instead we could do `label.single_lines.is_empty()`? This avoids iterating over all the lines in the diagnostic again.

codespan-reporting/src/term/views.rs
```diff
@@ -207,8 +211,21 @@ where
...
+                    {
+                        // Skip if there are too many lines in between
+                        continue;
```
brendanzab, line 226: Wondering if we could call `renderer.render_snippet_break(...)` here. 🤔

Johann150:
I redid it because it was not quite working how it was supposed to do, it was just not breaking the existing tests. I now consider it to be "working as intended", though not perfect yet.

Some things to consider:
- [ ] Should there be separate configuration options for context at start and end of the code snippet? Rust seems to have this, but it is probably just nice to have.
- [ ] What should be the default for how many lines are shown. Currently set to 3 like at the start of the Rust example in the Issue. But since it does not have as many lines of context at the end, maybe 2 would also be enough.

I added the two examples from the issue as tests which cover the errors I encountered in the process.

codespan-reporting/src/term/renderer.rs
```diff
@@ -23,7 +23,7 @@ pub type SingleLabel<'diagnostic> = (LabelStyle, Range<usize>, &'diagnostic str)
 
 /// A multi-line label to render.
 ///
-/// Locations are relative to the start of where the source cord is rendered.
+/// Locations are relative to the start of where the source code is rendered.
```
brendanzab, line 26: Heh, thanks for picking this up!

codespan-reporting/src/term/renderer.rs
```diff
@@ -42,7 +42,7 @@ pub enum MultiLabel<'diagnostic> {
     /// ```text
     /// │
     /// ```
-    Left,
+    Left(bool),
```
brendanzab, line 45: What is the `bool` meant to be here? Would be handy to have it explained in the doc comment. Boolean flags can be a little hard to figure out in general!

codespan-reporting/src/term/views.rs
```diff
@@ -207,8 +211,11 @@ where
                 // 6 │ │     0 _ => "Fizz"
                 // 7 │ │     _ 0 => "Buzz"
                 // ```
-                // TODO(#125): If start line and end line are too far apart, add a source break.
                 for line_index in (start_line_index + 1)..end_line_index {
+                    let optional =
+                        std::cmp::min(line_index - start_line_index, end_line_index - line_index)
+                            > renderer.context_lines();
```
brendanzab, line 217: Can we think of a better name for this `optional` identifier perhaps?

codespan-reporting/tests/snapshots/term__multiline_omit__rich_no_color.snap
```diff
@@ -0,0 +1,35 @@
...
+expression: test_data.emit_no_color(&config)
+---
+error[empty_if]: empty elseif block
+   ┌─ empty_if_comments.lua:1:1
+   │    
+ 1 │ ╭   elseif 3 then
+ 2 │ │   
+ 3 │ │ ╭ 
+ 4 │ │ │ 
+   · │ │
```
brendanzab, line 12: Yeah I like this rendering better - ie. continuing the lines over the source break. Good stuff 👍

codespan-reporting/tests/snapshots/term__multiline_omit__rich_color.snap
```diff
@@ -0,0 +1,56 @@
...
+{fg:Blue}21{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}cream chocolate cake tart. Brownie lemon drops sesame snaps. Oat cake bear claw {/}
+{fg:Blue}22{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}pudding macaroon bonbon. Danish tart cookie pastry cookie. Chocolate cake icing {/}
+   {fg:Blue}·{/}  
+{fg:Blue}27{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}roll croissant caramels tiramisu tart. Macaroon cupcake marshmallow toffee lemon {/}
+{fg:Blue}28{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}drops. Pie chocolate cake jelly beans candy canes icing pudding cheesecake marzipan. {/}
+{fg:Blue}29{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}Jelly liquorice apple pie fruitcake lemon drops candy gummi bears sugar plum. {/}
+{fg:Blue}30{/} {fg:Blue}│{/} {fg:Green}│{/} {fg:Green}Tootsie roll bonbon pastry toffee.{/}
+   {fg:Blue}│{/} {fg:Green}╰{/}{fg:Green}──────────────────────────────────^ this is the whole paragraph{/}
+   {fg:Blue}│{/}  
+   {fg:Blue}={/} this diagnostic will actually skip over some lines
```
brendanzab, line 54: This snapshot seems to have been leftover?

codespan-reporting/tests/term.rs
```diff
@@ -844,3 +844,77 @@ mod unicode_spans {
...
+                    Label::secondary(file_id2, 55..55).with_message("missing whitespace"),
+                ])
+                .with_notes(vec![
+                    "note:\texpected type `()`\n\tfound type `{integer}`".to_owned()
+                ]),
+        ];
+
+        let test_data = TestData { files, diagnostics };
+
+        insta::assert_snapshot!(test_data.emit_no_color(&config));
```
brendanzab, line 918: Would it be possible to use the same pattern as the other tests here? ie. doing some setup, and then using a macro to run the test?

codespan-reporting/src/term/views.rs
```diff
@@ -286,38 +293,72 @@ where
...
+                    (
+                        index,
+                        line,
+                        line.multi_labels.iter().any(|l| {
+                            if let (_, _, MultiLabel::Left(true)) = l {
+                                false
+                            } else {
+                                true
+                            }
+                        }) || !line.single_labels.is_empty(),
```
brendanzab, line 309: To be honest I'm finding it rather hard to understand what's going on here! Would be cool to see if we can figure out how to make this easier to understand, as the view/renderer code is already starting to get rather convoluted by my own hand. 😅

codespan-reporting/tests/term.rs
```diff
@@ -844,3 +844,77 @@ mod unicode_spans {
     test_emit!(rich_no_color);
     test_emit!(short_no_color);
 }
+
+mod multiline_omit {
```
brendanzab, line 848: Yay, a test, thanks! 🎉

brendanzab:
> Should there be separate configuration options for context at start and end of the code snippet? Rust seems to have this, but it is probably just nice to have.

What does Rust's configuration options look like?

I'm ok with not having this configurable - we can always add that in later if need be.

codespan-reporting/src/term/renderer.rs
Johann150: This bool indicates if this left label is "optional". The optionality is stored in the label. Then we can look at all the computed labels afterwards and see easily if the line can be skipped.

The previous approach of looking this up while looking at a single label would not work if there are mulitple `MultiLabel`s on one line which can both be skipped.

codespan-reporting/tests/snapshots/term__multiline_omit__rich_no_color.snap
Johann150: Noticed that you already thought ahead with `render_snippet_break`. I just had to put in the multilables on that line so it renders like that. 😄

codespan-reporting/src/term/renderer.rs
Johann150: I suppose the solution is to add a comment. Making it struct-like is probably too much.

Johann150:
I was probably not clear enough: Rust does not have a configuration option, but you would have to use the separate configuration options to realise rust-like diagnostics:

Rust shows 3 lines of context at the start and 1 line at the end of a snippet.

brendanzab:
Yeah, perhaps lets keep to what Rust does for now, then add a config if it's requested.

codespan-reporting/src/term/renderer.rs
brendanzab: Riiiight, so we make it `true` if it's in the three lines at the start, or one line of context at the end?

brendanzab: And otherwise `false` (it could overlap with another mandatory line)

brendanzab:
Yeah, wondering if we should split the config into `start_context_lines` and `end_context_lines` 🤔

codespan-reporting/src/term/renderer.rs
```diff
@@ -127,6 +129,10 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
         &self.config.styles
     }
 
+    pub fn context_lines(&self) -> usize {
+        self.config.context_lines
+    }
```
brendanzab, line 134: I'm guessing this is redundant now?

codespan-reporting/src/term/renderer.rs
brendanzab: After some thought, I'm wondering if it would be cleaner/less confusing if we put this flag on the `Line` struct in the view. And we update it as we add new labels in the initial pass over the diagnostic.

codespan-reporting/src/term/renderer.rs
Johann150: No, this is the only way to access the configuration from within `RichDiagnostic::render`.

brendanzab: Ahhh, interesting! Yeah it seems kind of annoying that we need to propagate information back to the view like this. Not sure what can be done about it though, and the separation between the view and the renderer is not the greatest right now if I'm honest - ie. we do a bit of layout-ish computation in the renderer.

Johann150:
I think there is still one point that should be thought about: *This feature cannot be disabled.* I think setting both context settings to `std::usize::MAX` should practically disable it for most applications. It might even be a good idea to limit the number of lines, even if this limit would never realistically be reached.

codespan-reporting/src/term/renderer.rs
Johann150: But I think this problem is outside the scope of this PR, right?

brendanzab: Yeah, agreed. Sorry I should have been more clear on that. 😓

codespan-reporting/src/term/views.rs
```diff
@@ -200,24 +211,35 @@ where
...
-                    labeled_file
-                        .get_or_insert_line(line_index, line_range, line_number)
-                        .multi_labels
+                    let line = labeled_file.get_or_insert_line(line_index, line_range, line_number);
+
+                    line.multi_labels
                         .push((label_index, label.style, MultiLabel::Left));
+
+                    // The line should be rendered to match the configuration of how much context to show.
+                    line.omittable &= !(
```
brendanzab, line 236: I'm wondering if this would be clearer if the flag was reversed, eg. calling it `must_render` or something 🤔

codespan-reporting/src/term/views.rs
```diff
@@ -303,21 +335,31 @@ where
...
+                            // get filtered out labels, if there are any
+                            let labels = if let Some((_, line)) = labeled_file
+                                .lines
+                                .iter()
+                                .find(|(index, _)| **index == line_index + 1)
+                            {
+                                &line.multi_labels
+                            } else {
+                                &current_labels
+                            };
```
brendanzab, line 354: Not really understanding what this line is for. What happens if we just use `&current_labels` in the call to `render_snippet_source` instead of doing this?

Johann150: This is inside a match case that is invoked if the next line that must be rendered is 2 lines away. In this case it does not make sense to hide the next line (making a one line omission) and it is rendered although it was previously filtered out. Therefore the labels on that line have to be retrieved again from the unfiltered `labeled_file.lines`.
Although I must admit, it would probably be cleaner with a `map_or`, I'll fix that.

codespan-reporting/src/term/views.rs
```diff
@@ -286,7 +313,11 @@ where
                 )?;
             }
 
-            let mut lines = labeled_file.lines.into_iter().peekable();
+            let mut lines = labeled_file
+                .lines
+                .iter()
+                .filter(|(_, line)| line.must_render)
+                .peekable();
             let current_labels = Vec::new();
```
brendanzab, line 321: Ahh! I think I intended `current_labels` to be a stack originally. I should have put a `TODO` on this. Perhaps it can be removed now? The other usages can now be replaced with `&[]`, which is a bit more illustrative, I think. I can do this in another PR if you like, I know this has taken a while to review, so I'm happy to merge now if you like to get it off your hands.

codespan-reporting/src/term/views.rs
```diff
@@ -303,21 +334,30 @@ where
...
                             // Write a source line
                             let file_id = labeled_file.file_id;
+
+                            // This line was not intended to be rendered initially.
+                            // To render the line right, we have to get back the original labels.
+                            let labels = labeled_file
+                                .lines
+                                .iter()
+                                .find(|(index, _)| **index == line_index + 1)
+                                .map_or(&current_labels, |(_, line)| &line.multi_labels);
```
brendanzab, line 351: I just noticed that `labeled_file.lines` is a `BTreeMap`, so we could probably do something like:

```rust
let labels = labeled_file
    .lines
    .get(line_index + 1)
    .map_or(&[], |(_, line)| &line.multi_labels);
```

codespan-reporting/src/term/views.rs
```diff
@@ -346,9 +345,10 @@ where
                             // To render the line right, we have to get back the original labels.
                             let labels = labeled_file
                                 .lines
-                                .iter()
-                                .find(|(index, _)| **index == line_index + 1)
-                                .map_or(&current_labels, |(_, line)| &line.multi_labels);
+                                .get(&(line_index + 1))
+                                .map_or(&[] as &[(usize, LabelStyle, MultiLabel)], |line| {
```
Johann150, line 349: I don't know if I did something wrong, but Rust would not accept just `&[]` and would error on the map function with: `expected array of 0 elements, found slice`.

brendanzab: Ahh yeah that is annoying - I think I've done `&[][..]` in the past to fix this

codespan-reporting/src/term/renderer.rs
```diff
@@ -127,6 +127,14 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
         &self.config.styles
     }
 
+    pub fn start_context_lines(&self) -> usize {
+        self.config.start_context_lines
+    }
+
+    pub fn end_context_lines(&self) -> usize {
+        self.config.end_context_lines
+    }
```
brendanzab, line 136: Let's add a comment here, just so I remember:
```suggestion
    // FIXME: Architectural smell - we'd prefer not to pass information from the renderer to the view.
    pub fn start_context_lines(&self) -> usize {
        self.config.start_context_lines
    }

    // FIXME: Architectural smell - we'd prefer not to pass information from the renderer to the view.
    pub fn end_context_lines(&self) -> usize {
        self.config.end_context_lines
    }
```

brendanzab:
Thanks so much for bearing with me on the review! This is great! :)

Johann150:
No problem. 👍

Issue #227, opened 2020-04-06
codespan-reporting and codespan-lsp compatibility


aweary:
I've been using `codespan` and `codespan-reporting` for error reporting for my CLI. I'm not working on an LSP extension and want to use `codespan-lsp` to covert to the proper LSP types, but it seems like `codespan-reporting` doesn't play well.

For example, I have a custom file system abstraction that implements [the `Files` trait](https://docs.rs/codespan-reporting/0.9.2/codespan_reporting/files/trait.Files.html) but `codespan-lsp` utilities like [`byte-span-to-range`](https://docs.rs/codespan-lsp/0.9.2/codespan_lsp/fn.byte_span_to_range.html) expect a reference to [`codespan::Files`](https://docs.rs/codespan/0.9.2/codespan/struct.Files.html). They also expect a [`Span`](https://docs.rs/codespan/0.9.2/codespan/struct.Span.html) but `codespan-reporting`'s [`Label`](https://docs.rs/codespan-reporting/0.9.2/codespan_reporting/diagnostic/struct.Label.html) uses `Range<usize>`.

Are there plans to align the packages one way or the other, or should I just roll my own diagnostics object and have compatibility layers for `codespan-reporting` and `codespan-lsp`?

brendanzab:
Thanks for your issue!

I've been trying hard to de-emphasise `codespan`, and emphasizing `codespan-reporting`, because I've found it rather challenging to create a one-size-fits all thing for handling files. This was part of the work I did in inverting the dependency between those two crates. Eventually I'd like to deprecate `codespan`, if I'm honest.

`codespan-lsp` currently depends on `codespan` though, so sadly you are tied to using `codespan` if you want these conversions. It would be cool if there would be some way to make those conversions easier to implement for custom file system abstractions though! I'm still not sure where they fit though…

aweary:
@brendanzab would you be open to including some LSP utilities in `codespan-reporting`? It might make sense to have `codespan_reporting::lsp` to mirror the `codespan_reporting::term` backend. I'd be happy to help here if you're interested.

brendanzab:
Ohhh, that is a good idea, yes! Now I have a better handle on the `lsp-types` version ranges this could also be helpful too - it was a constant pain to keep having to update them.

brendanzab:
I'd definitely be open to collaborating on this! Feel free to chat on [Gitter](https://gitter.im/codespan-rs/Lobby) or the the #langdev channel on the rust community discord if you like!

PR #258, opened 2020-07-05
feat: Make codespan_lsp use codespan_reporting

Marwes:
Makes `codespan_lsp` more broadly usable by lifting the `codespan`
restriction. `codespan` users can still used it after converting to and
from usize for arguments and return values.

BREAKING CHANGE

The signatures now take usize instead of `codespan`s explicit, byte,
line, column types.

The error types have been copied into codespan_lsp to remove the
dependency on codespan.

### DIFF
--- a/codespan-lsp/src/lib.rs
+++ b/codespan-lsp/src/lib.rs
@@ -249,7 +249,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use codespan::{Files, Location};
+    use codespan_reporting::files::{Location, SimpleFiles};
 
     use super::*;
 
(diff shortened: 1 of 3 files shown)


codespan-lsp/src/lib.rs
```diff
@@ -140,38 +211,45 @@ pub fn character_to_line_offset(line: &str, character: u64) -> Result<ByteOffset
...
+    F: Files<'a> + ?Sized,
+{
+    Ok(position_to_byte_index(files, file_id, &range.start)?
+        ..position_to_byte_index(files, file_id, &range.end)?)
 }
 
 #[cfg(test)]
 mod tests {
-    use codespan::Location;
+    use codespan::{Files, Location};
```
brendanzab, line 252: Maybe we could use `codespan_reporting::SimpleFiles` for this?

brendanzab:
Oh nice! I wonder how much this relates to #227 🤔

Marwes:
cc @aweary ?

PR #260, opened 2020-07-08
locus at first primary label

### LINKED ISSUE #259: File column/line should consistently point to the first primary label instead of the first label
Currently when you have a diagnostic, and that diagnostic contains 2 labels, say, one secondary at `1:1` and one primary at `6:3`, the column/line at the top of the diagnostic on rich rendering shows `1:1`, this should instead point to the first primary label, which seems to be the case for short rendering. When you use short rendering it points to `6:3`, this should be made consistent across rendering types.

Johann150:
This should resolve #259.

This solution only works if the first primary label occurs before any secondary labels which appear earlier in the source code. This would e.g. also work if all primary labels were added before the secondary labels. I think that is more or less the convention? 

If this behaviour is adopted, it should be documented somewhere. Where would that best be?

Alternatively the labels could be sorted before being processed. But I think that might be a performance hit.

### DIFF
--- a/codespan-reporting/assets/readme_preview.svg
+++ b/codespan-reporting/assets/readme_preview.svg
@@ -52,23 +52,23 @@
   <foreignObject x="0" y="0" width="882" height="335">
     <div xmlns="http://www.w3.org/1999/xhtml">
       <pre><span class="fg red bold bright">error[E0308]</span><span class="bold bright">: `case` clauses have incompatible types</span>
-   <span class="fg cyan">┌─</span> FizzBuzz.fun:16:16
-   <span class="fg cyan">│</span>  
-<span class="fg cyan">10</span> <span class="fg cyan">│</span>   fizz₂ : Nat → String
-   <span class="fg cyan">│</span>                 <span class="fg cyan">------</span> <span class="fg cyan">expected type `String` found here</span>
-<span class="fg cyan">11</span> <span class="fg cyan">│</span>   fizz₂ num =
-<span class="fg cyan">12</span> <span class="fg cyan">│</span> <span class="fg cyan">╭</span>     case (mod num 5) (mod num 3) of
-<span class="fg cyan">13</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         0 0 =&gt; "FizzBuzz"
-   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg cyan">----------</span> <span class="fg cyan">this is found to be of type `String`</span>
-<span class="fg cyan">14</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         0 _ =&gt; "Fizz"
-   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg cyan">------</span> <span class="fg cyan">this is found to be of type `String`</span>
-<span class="fg cyan">15</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         _ 0 =&gt; "Buzz"
-   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg cyan">------</span> <span class="fg cyan">this is found to be of type `String`</span>
-<span class="fg cyan">16</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         _ _ =&gt; <span class="fg red">num</span>
-   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg red">^^^</span> <span class="fg red">expected `String`, found `Nat`</span>
-   <span class="fg cyan">│</span> <span class="fg cyan">╰</span><span class="fg cyan">──────────────────' `case` clauses have incompatible types</span>
-   <span class="fg cyan">│</span>  
-   <span class="fg cyan">=</span> expected type `String`
+   <span class="fg blue">┌─</span> FizzBuzz.fun:16:16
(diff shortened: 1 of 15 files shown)


codespan-reporting/src/term/views.rs
```diff
@@ -100,14 +100,18 @@ where
...
                         labeled_file.location =
                             files.location(label.file_id, label.range.start).unwrap();
                     }
                     labeled_file
                 }
                 None => {
+                    // no other diagnostic referenced this file yet
                     labeled_files.push(LabeledFile {
                         file_id: label.file_id,
                         start: label.range.start,
```
Johann150, line 117: To make the solution truly universal, it would have to be made sure that this case is first triggered by a primary label.

Or LabeledFile::location could also be an `Optional` so it is only set to `Some` when a primary label is found and the positions of secondary labels would be ignored. This would add an error condition.

brendanzab: I wonder if instead we could have a `max_label_style: LabelStyle` field on `LabeledFile`? This would let us fall-back to the first secondary label, and then as soon as we see the first primary label we'd be able to switch to using that as our start position.

Johann150:
I think this is ready to be merged.

codespan-reporting/src/term/views.rs
```diff
@@ -100,21 +101,27 @@ where
...
+                    // another diagnostic also referenced this file
+                    if labeled_file.start > label.range.start
+                        && labeled_file.max_label_style >= label.style
+                    {
+                        // this label indicates an earlier start
                         labeled_file.start = label.range.start;
                         labeled_file.location =
                             files.location(label.file_id, label.range.start).unwrap();
                     }
                     labeled_file
```
brendanzab, line 113: Why do we not update `max_label_style` here?

codespan-reporting/assets/readme_preview.svg
```diff
@@ -52,23 +52,23 @@
...
+<span class="fg cyan">14</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         0 _ =&gt; "Fizz"
+   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg cyan">------</span> <span class="fg cyan">this is found to be of type `String`</span>
+<span class="fg cyan">15</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         _ 0 =&gt; "Buzz"
+   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg cyan">------</span> <span class="fg cyan">this is found to be of type `String`</span>
+<span class="fg cyan">16</span> <span class="fg cyan">│</span> <span class="fg cyan">│</span>         _ _ =&gt; <span class="fg red">num</span>
+   <span class="fg cyan">│</span> <span class="fg cyan">│</span>                <span class="fg red">^^^</span> <span class="fg red">expected `String`, found `Nat`</span>
+   <span class="fg cyan">│</span> <span class="fg cyan">╰</span><span class="fg cyan">──────────────────' `case` clauses have incompatible types</span>
+   <span class="fg cyan">│</span>  
+   <span class="fg cyan">=</span> expected type `String`
```
brendanzab, line 71: Woops, seems like this is being rendered differently on Windows. We need to update the svg example to a platform-consistent configuration (as we do in the integration tests)

Johann150: Should I add that to this PR? Seems quite easy to do.

brendanzab: That would be great, if you can! 🤩

brendanzab:
Would be cool if we could rebase this branch now to a single commit (or whatever granularity makes sense). That way the history is a bit nicer. :)

Johann150:
The checks should actually pass, I changed the Rustfmt issue in the last commit. Maybe because I force-pushed...

> Would be cool if we could rebase this branch now to a single commit (or whatever granularity makes sense). That way the history is a bit nicer. :)

Is that something I have to do?

brendanzab:
> Pull Request should allow for sqashing commits.

Ahhh I often do an interactive rebase to group things up into commits that each pass the test suite (if multiple commits make things easier to understand), but I can just use squash-and-merge for this.

brendanzab:
I'll try re-running the jobs to see if it passes :)

codespan-reporting/src/diagnostic.rs
```diff
@@ -112,6 +112,8 @@ impl<FileId> Label<FileId> {
 
 /// Represents a diagnostic message that can provide information like errors and
 /// warnings to the user.
+///
+/// The position of a Diagnostic is considered to be the position of the first [`Label`] with a style of [`LabelStyle::primary`].
```
brendanzab, line 116: Do we need this comment anymore, seeing as we now get the primary label with the smallest start position?

Johann150: That is what I was trying to say: The Position of the first label with a style of primary, so the first primary label. But I see how "first" vs. "smallest start position" is confusing. Maybe like this:

```suggestion
/// The position of a Diagnostic is considered to be the position of the [`Label`] with a style of [`LabelStyle::primary`] that has the smallest start position.
```

Or should it be removed completely? I just thought that it would be good to document this behaviour.

brendanzab: Ahh yeah, I'm not yet decided on if we want to talk about presentation stuff here - but on the other hand, that's where people will probably look first. Let's commit your rewording. :+1:

PR #262, opened 2020-07-12
fix: Handle unicode in codespan_lsp outside of the first line

Marwes:


codespan-lsp/src/lib.rs
```diff
@@ -229,8 +229,9 @@ where
     let source = source.as_ref();
 
     let line_span = files.line_range(file_id, position.line as usize).unwrap();
+    let line_str = source.get(line_span.clone()).unwrap();
 
-    let byte_offset = character_to_line_offset(source, position.character)?;
+    let byte_offset = character_to_line_offset(line_str, position.character)?;
```
brendanzab, line 234: What was the original issue with this?

brendanzab: Ohhh, it was getting the character offset in the whole file, then adding it to the line start position.

PR #263, opened 2020-07-12
no excessive highlighting

### LINKED ISSUE #261: There should be no source color highlighting if the label points to the whole line, or if the label is multiline
Codespan currently highlights in red whatever the primary label is pointing to in the source, while this looks great for errors inside of source code lines, it looks really ugly if the label points to the whole line. It ruins the point of highlighting, you can clearly see what the label is pointing at, a line of all red just looks like it does not belong there. In the case of multiline labels, usually they are used to point to blocks of source code, but this means if you highlight everything inside of them you end up with a giant blob of RED which just looks like it does not belong and it ruins the cleanliness of the diagnostic.

Labels should highlight their source code only if the label points to a subsection of the source code line, and should be fully disabled for multiline labels.

Johann150:
This should resolve #261.

Is this just okay like this, or should this be configurable?

### DIFF
--- a/codespan-reporting/src/term/renderer.rs
+++ b/codespan-reporting/src/term/renderer.rs
@@ -244,6 +244,27 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
         // FIXME: Use the number of trimmed placeholders when rendering single line carets
         let source = source.trim_end_matches(['\n', '\r', '\0'].as_ref());
 
+        // count the number of whitespace characters
+        let text_start = source.chars().take_while(|c| c.is_whitespace()).count();
+
+        // get the byte index of the first non-whitespace character
+        let text_start = source
+            .char_indices()
+            .nth(text_start)
+            .map(|(index, _)| index)
+            .unwrap_or(0);
+
+        // get the byte index of the first trailing whitespace character
+        let text_end = source
+            .char_indices()
+            .rev()
+            .take_while(|(_, c)| c.is_whitespace())
+            .last()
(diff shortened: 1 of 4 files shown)


brendanzab:
Sorry for my slow review on this - need to figure out whether I'm happy with how we're iterating over the characters of a source line.

codespan-reporting/src/term/renderer.rs
```diff
@@ -240,6 +240,20 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+        let text_start = source.find(|c| !char::is_whitespace(c)).unwrap_or(0);
+
+        // get the byte index of the first trailing whitespace character
+        let text_end = source
+            .char_indices()
+            .rev()
+            .take_while(|(_, c)| c.is_whitespace())
+            .last()
+            .map(|(index, _)| index)
+            .unwrap_or_else(|| source.len());
```
Johann150, line 253: This could also be rewritten using str::rfind, but because we have to find the first trailing non-whitespace character, we would have to find the char lengt of the character to adjust, which makes it not much clearer than the curren solution.
```suggestion
        let text_end = source
            .rfind(|c| !char::is_whitespace(c))
            .map(|index|
                index + source[index..]
                    .chars()
                    .next()
                    .map_or(0, |c| c.len_utf8())
            )
            .unwrap_or_else(|| source.len());
```

Johann150:
@brendanzab any feedback on this?

Issue #85, opened 2019-06-10
Add simple diagnostic rendering mode


brendanzab:
As part of #30 it might be nice to have a simple diagnostic rendering mode to make it easier on screen readers. This would only display a list of diagnostics, without the source snippets.

Here's an example of what it might look like:

```
error[E0308]: `case` clauses have incompatible types
    - FizzBuzz.fun:8:12: expected `String`, found `Nat`
        = expected type `String`
             found type `Nat`
        = FizzBuzz.fun:4:13: `case` clauses have incompatible types
        = FizzBuzz.fun:3:15: expected type `String` found here

error[E0308]: `case` clauses have incompatible types
    - FizzBuzz.fun:15:16: expected `String`, found `Nat`
        = expected type `String`
             found type `Nat`
        = FizzBuzz.fun:11:5: `case` clauses have incompatible types
        = FizzBuzz.fun:12:16: this is found to be of type `String`
        = FizzBuzz.fun:13:16: this is found to be of type `String`
        = FizzBuzz.fun:14:16: this is found to be of type `String`
```

Perhaps others have better ideas though?

brendanzab:
Rustc has a `--message-format short` mode too, which effectively looks like:

```
FizzBuzz.fun:8:12: error[E0308]: `case` clauses have incompatible types
FizzBuzz.fun:15:16: error[E0308]: `case` clauses have incompatible types
```

Johann150:
Combining with information from #30, I think maybe a "medium mode" where notes are also rendered might be a good idea.

Another Idea would be to write information so it is easier to understand when only read by a screen reader (which might not pronounce punctuation characters for example). But this is probably something that would better be done using the proposed JSON output.

As an example, one could render the first example from the `fizz_buzz` test like this:
```
error E0308: `case` clauses have incompatible types
in file FizzBuzz.fun on line 8, column 12.
additional information:
line 8, column 12: expected `String`, found `Nat`.
line 3, column 15: expected type `String` found here.
line 4, column 14 until line 8: `case` clauses have incompatible types.
```

brendanzab:
That sounds _really_ cool! I like it! It would be nice if we offered the screen-reader support out of the box, but then allow people to build their own tools using JSON.

Johann150:
[...]

PR #266, opened 2020-07-15
medium rendering mode

Johann150:
This is a suggestion from #85.

Just added a bit onto the short mode rendering because otherwise big chunks of code would have to be duplicated.

Tests still have to be added.

### DIFF
--- a/codespan-reporting/src/term/views.rs
+++ b/codespan-reporting/src/term/views.rs
@@ -419,6 +419,7 @@ where
 /// Output a short diagnostic, with a line number, severity, and message.
 pub struct ShortDiagnostic<'diagnostic, FileId> {
     diagnostic: &'diagnostic Diagnostic<FileId>,
+    show_notes: bool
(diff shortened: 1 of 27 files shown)


Johann150:
I added the tests for medium mode almost everywhere. If that's too much (total number of tests increased from 54 to 77), I can also remove some of them again.

codespan-reporting/src/term/views.rs
```diff
@@ -454,6 +454,7 @@ where
 /// Output a short diagnostic, with a line number, severity, and message.
 pub struct ShortDiagnostic<'diagnostic, FileId> {
     diagnostic: &'diagnostic Diagnostic<FileId>,
+    show_notes: bool,
```
brendanzab, line 457: On a later read of this I'm not really happy with the coupling of these two views, but I think that can be fixed in a later PR. Let's merge this - it's been open for too long!

PR #269, opened 2020-07-21
Remove Files trait, switch to a more dynamic file fetching system

zesterer:
This PR is only partially complete. In theory is removes a lot of the complexity surrounding the `Files` trait while also making the API more useful for users. Right now, `Files` has not been removed and as such, a lot of extra code is used to allow compatibility between both the old and new system. This will be removed in time and complexity shall be significantly reduced.

### DIFF
--- a/codespan-reporting/examples/custom_files.rs
+++ b/codespan-reporting/examples/custom_files.rs
@@ -33,7 +33,7 @@ fn main() -> anyhow::Result<()> {
     let config = codespan_reporting::term::Config::default();
     for message in &messages {
         let writer = &mut writer.lock();
-        term::emit(writer, &config, &files, &message.to_diagnostic())?;
+        term::emit(writer, &config, |id| files.get(id).unwrap(), &message.to_diagnostic())?;
     }
 
     Ok(())
@@ -46,7 +46,7 @@ mod files {
 
     /// A file that is backed by an `Arc<String>`.
     #[derive(Debug, Clone)]
-    struct File {
+    pub struct File {
         /// The name of the file.
         name: String,
         /// The source code of the file.
@@ -65,6 +65,12 @@ mod files {
                 Ordering::Greater => None,
             }
         }
+    }
+
+    impl files::Src for File {
+        fn name(&self) -> &str { &self.name }
+        fn source(&self) -> &str { &self.source }
+        fn line_starts(&self) -> Option<&[usize]> { Some(&self.line_starts) }
     }
 
     /// An opaque file identifier.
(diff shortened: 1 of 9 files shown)


Johann150:
As familiar as I am with codespan-reporting (which is not very deeply), I think there is a reason why the `Files` trait exists: It allows dependants to redefine this system.

If you are not satisified with `SimpleFiles`, you could make your own struct to replace it, if it also implements the `Files` trait.

zesterer:
@Johann150 I've discussed the shortcomings of the `Files` trait with @brendanzab via Discord. The lack of `&mut` on methods requires the API user to do interior mutability if they want to implement any sort of cache: however, this conflicts directly with the `AsRef<str>` requirement of fields (`cell::Ref` does not implement `AsRef<Inner>`). This then requires writing a wrapper type for `cell::Ref`, massively inflating the complexity of the boilerplate needed to properly glue the `Files` trait to anything that requires mutability.

This PR, though incomplete, simplifies this situation significantly by having `term::emit` take an `FnMut`. This allows the API user to use mutation if they want while not harming those that do not. It also cleans up the `Files` trait by simplifying the requirements and making it easier for both newcomers to interact with (via internal caching of `line_starts`) and for more advanced users to provide such a cache (all that is necessary is to override the default method impl in `Src`).

Johann150:
I hope this is not just going over my head, but I think that taking a closure is messier than the way it was before. But that might just be personal preference.

On another note: IMHO there is a good reason why the file database should be immutable (except for inserting files): If you change or remove a file from the database, this likely throws off all Diagnostics currently pointing at that file, thus being an easy recipe for footshoting. Maybe you could explain a bit why you (or brendanzab) think this should change. Or give a use case for mutability, because I cannot think of one that is better solved another way.

zesterer:
@Johann150 Having the file database be mutable allows for caching. It's common that I want to pull up a snippet from a file that's part of a previous compilation unit without needing to pull it up in advance.

Johann150:
So @zesterer and I talked this through elsewhere because I did not quite get the idea behind it. The gist I got is this:

Consider caching compilation units which are compiled separately. The renderer might or might not access the content of files compiled previously. To minimize which files have to be loaded into memory, a files content would only be loaded once the renderer tries to access it. To detect this, the caching would have to be implemented inside of a file database which implements the `Files` trait. This could then load a file upon the first call to e.g. `source` using the respective file id (thus mutating `self`). But that is impossible because the required methods of `Files` accept `& self`. But just changing the methods to accept `&mut self` might also not be desirable in some cases. Using `FnMut` resolves this because it can handle both cases.

brendanzab:
Going to close this for now! Might be worth making an issue about these design decisions if you are still interested.

PR #274, opened 2020-08-02
Add the diagnostic! macro

### LINKED ISSUE #273: Introduce macro to create diagnostics
It would be nice to have a macro that can create a `Diagnostic`.
Currently, the code to create a `Diagnostic` can be very clumsy and messy.
E.g.
```rust
let d = Diagnostic::error();
let label = Label::primary(file, span)
    .with_message(format!("expected '{}', found '{}'", expected, found));
d.with_message("unexpected token")
    .with_labels(vec![label])
    .with_code("E123")
    .with_notes(vec!["this token is wrong".to_string()])
```
A good macro could replace the code above with something like this:
```rust
diagnostic! {
    error => "unexpected token",
    label: primary(format!("expected '{}', found '{}'", expected, found), file, span),
    code: "E123",
    note: "this token is wrong",
}
```

Kixiron:
Adds a macro to create `Diagnostic`s with

Overblown usage to show everything that the macro can do:

```rust
// Invocation
diagnostic! {
    error: "unexpected token",
    label: Label::primary(format!("expected '{}', found '{}'", expected, found), file, span)),
    labels: [Label::secondary("You could make it right by changing it to `tok`")],
    code: "E123",
    note: "this token is wrong",
    notes: ["I", "ran", "out", "of", "note", "ideas"]
}

// Creates the Diagnostic
Diagnostic {
    message: "unexpected token",
    severity: Severity::Error,
    labels: [
        Label::primary(format!("expected '{}', found '{}'", expected, found), file, span)),
        Label::secondary("You could make it right by changing it to `tok`"),
    ],
    notes: [
        "this token is wrong",
        "I", "ran", "out", "of", "note", "ideas",
    ],
    code: Some("E123"),
}
```

Normal usage that doesn't make your head hurt:

```rust
// Invocation
diagnostic! {
    error: "'error' detected?",
    label: Label::primary("This text has been detected as being 'incorrect'... or something".to_owned()),
    note: "I'm not 100% sure of very much, but there could possibly be something wrong",
}

// Creates the Diagnostic
Diagnostic {
    message: "'error' detected?",
    severity: Severity::Error,
    labels: [Label::primary("This text has been detected as being 'incorrect'... or something")],
    notes: ["I'm not 100% sure of very much, but there could possibly be something wrong"],
    code: None,
}
```

Closes #273

### DIFF
--- a/codespan-reporting/CHANGELOG.md
+++ b/codespan-reporting/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added the `diagnostic!` macro
+
+### Changed
+
+- Added `PartialEq` implementations to `Diagnostic` and `Label` where `FileId` implements `PartialEq`
+
 ## [0.9.5] - 2020-06-24
 
 ### Changed
--- a/codespan-reporting/src/diagnostic.rs
+++ b/codespan-reporting/src/diagnostic.rs
@@ -60,7 +60,7 @@ pub enum LabelStyle {
 }
 
 /// A label describing an underlined region of code associated with a diagnostic.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub struct Label<FileId> {
     /// The style of the label.
@@ -114,7 +114,7 @@ impl<FileId> Label<FileId> {
 /// warnings to the user.
 ///
 /// The position of a Diagnostic is considered to be the position of the [`Label`] with a style of [`LabelStyle::primary`] that has the smallest start position.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "serialization", derive(Serialize, Deserialize))]
 pub struct Diagnostic<FileId> {
     /// The overall severity of the diagnostic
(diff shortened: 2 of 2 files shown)


brendanzab:
Hey, thanks for the contribution! And sorry for the late response!

I think on the whole I'm still a bit on the fence on this - like, I'm not sure the case has been made on the original issue yet for its inclusion. I'm also a bit wary of the special cases (like `label` vs. `labels`). I do appreciate the fact that there are some good docs to go along with this though!

Kixiron:
The special casing of `label`/`labels` and `note`/`notes` is purely for convenience, since it'll allow single-element usages to have a little less visual noise. I'm not very tied to it, I'd be happy to remove them if you'd like

I think it's a nice addition personally, it gives some nice sugar to creating diagnostics

Johann150:
Adding to what brendanzab said is that this macro (at least to me) seems to be quite complicated and thus might be difficult to maintain. If we choose to add this macro (which I do not suggest), it should be simplified a bit. This is the bigger scale of the the issue brendanzab was pointing out with the `label`/`labels` example.

brendanzab:
Tbh I think the complexity is fine for the syntax it is defining - it's a recusive macro, and that's kind of needed for this kind of thing (ie. defining a struct-style of syntax using `macro_rules!`). Not sure if there's any way around it. I guess I'm more interested if there are other ways to improve the ergonomics of making diagnostics without needing to resort to macros.

Johann150:
As no other arguments have been put forth, it is not convincing that this is really necessary.

PR #275, opened 2020-08-14
errors instead of panics

### LINKED ISSUE #189: Return error values rather than panicking when rendering diagnostics.
We have lots of `unwraps` and fallible slice accesses in the diagnostic rendering at the moment that could result in user-facing panics when calling `codespan_reporting::term::emit`. Most of these panicking calls are currently localised to [`codespan_reporting::term::views`](https://github.com/brendanzab/codespan/blob/master/codespan-reporting/src/term/views.rs).
The panics could result from invalid file IDs, or invalid line ranges, say if an incremental compiler updated the underlying file before an error has been printed.

What should we do in this case? Should we panic? Return a richer error enum than a `std::io::Error`? Render something to the console?

Related: #140

Johann150:
resolves #189 

This is just a first idea of how it could look like. One point right away to consider is when some of these unwraps might fail and if the errors should be more specific then.

Something else is also that `std::io::Error` has to be wrapped. Maybe there is a nicer solution for that.

### DIFF
--- a/codespan-reporting/src/term.rs
+++ b/codespan-reporting/src/term.rs
@@ -28,6 +28,21 @@ impl From<std::io::Error> for RenderError {
     }
 }
 
+impl std::fmt::Display for RenderError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f,"{:?}",self)
+    }
+}
+
+impl std::error::Error for RenderError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match &self{
+            RenderError::IO(e)=>Some(e),
+            _=>None,
+        }
+    }
+}
+
 /// A command line argument that configures the coloring of the output.
 ///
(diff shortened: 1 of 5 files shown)


codespan-reporting/src/term.rs
```diff
@@ -15,6 +14,40 @@ pub use termcolor;
...
+        RenderError::Io(err)
+    }
+}
+
+impl std::fmt::Display for RenderError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RenderError::FileMissing => write!(f, "file missing"),
+            RenderError::InvalidIndex => write!(f, "invalid index"),
+            RenderError::Io(err) => write!(f, "{:?}", err),
```
Marwes, line 37:
```suggestion
            RenderError::Io(err) => write!(f, "{}", err),
```

brendanzab:
Ok, merged this! We might want to add some more descriptive information later, but I think this is good.

PR #279, opened 2020-08-31
remove indirection for configuration access

Johann150:
This removes the architectural smell of accessing the configuration for the view through the renderer.

The chosen solution is to hand another reference of the `Config` struct to the view.

I only added this onto `RichDiagnostic` because only it accesses the configuration and the compiler complains if a struct has an unused field.


Johann150:
As ratmice pointed out in the matrix chat, this is not a breaking change.

brendanzab:
Oooh looks good, thanks!

Yeah that was the original intention behind keeping this stuff private, while I figured it out. 😅

PR #280, opened 2020-09-02
Fix build on wasm with `serialization` feature

andrewhickman:
`serde` only implements `Serialize` for `OsString` on unix and windows, so this didn't compile previously.

This PR just does the most minimal change of disabling the `Serialize` impl on wasm.

### DIFF
--- a/codespan/src/file.rs
+++ b/codespan/src/file.rs
@@ -326,7 +326,11 @@ where
 
 /// A file that is stored in the database.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


brendanzab:
Thanks for the comment and the PR description! Might also be helpful to have a comment on the annotation saying why we are doing this, for future readers. :)

andrewhickman:
I've added a comment :)

brendanzab:
Thanks!

PR #282, opened 2020-09-04
fix locus calculation

### LINKED ISSUE #259: File column/line should consistently point to the first primary label instead of the first label

Johann150:
Changed the multiline_overlap test to cover this case too (and documented why the labels are ordered differently for the future). The output should be unchanged although the ordering of the labels in the struct has been changed.

This should resolve #259 again.


brendanzab:
Thanks!

PR #283, opened 2020-09-06
Do not render diagnostic code if it is an empty string

RDambrosio016:
Pretty much what the title says 😄 
If the code is some but it is empty then codespan will currently still render it, which results in an ugly `error[]`, this fixes that.

### DIFF
--- a/codespan-reporting/src/term/renderer.rs
+++ b/codespan-reporting/src/term/renderer.rs
@@ -168,7 +168,7 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
         // ```text
         // [E0001]
         // ```
-        if let Some(code) = &code {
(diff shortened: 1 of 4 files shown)


Johann150:
Oh thanks, yet another edge case found!
It would be nice if you added some tests to cover this. I think having `short_no_color` and `rich_no_color` tests would be enough for this case (i.e. no need for tests with colour).
If you don't know how to add tests, feel free to ask me for help.

Johann150 (approved):
Looks good. Just as an idea, using `and_then` instead of the `if` might be better?

RDambrosio016:
I generally try to avoid and_then for simple cases like this, the if (in my opinion) makes the code more understandable

brendanzab:
Thanks!

PR #287, opened 2020-09-15
move handling of label start to renderer (TODO comment)

Johann150:
This eliminates a failure condition from `RichDiagnostic::render` because the full source code does not have to be loaded here any more.
This also allows to remove the special `TopLeft` label style from the renderer.

I am not sure if this is actually an improvement or if this complicates the renderer more so see this more as how it could look.

### DIFF
--- a/codespan-reporting/src/term/renderer.rs
+++ b/codespan-reporting/src/term/renderer.rs
@@ -1,5 +1,5 @@
 use std::io::{self, Write};
-use std::ops::{Range, RangeTo};
+use std::ops::Range;
 use termcolor::{ColorSpec, WriteColor};
 
 use crate::diagnostic::{LabelStyle, Severity};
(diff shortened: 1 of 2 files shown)


codespan-reporting/src/term/renderer.rs
```diff
@@ -25,17 +25,18 @@ pub type SingleLabel<'diagnostic> = (LabelStyle, Range<usize>, &'diagnostic str)
...
     /// ╭────────────^
     /// ```
+    ///
+    /// Can also be rendered at the beginning of the line
+    /// if there is only whitespace before the label starts.
+    ///
+    /// /// ```text
+    /// ╭
+    /// ```
     Top(RangeTo<usize>),
```
Johann150, line 40: Is there any particular Reason why this is a `RangeTo` and not just a `usize`?

brendanzab: No reason - I think I was trying to communicate that it was the range from the start of the line. We could switch it to a `usize` if that's easier though.

brendanzab (approved):
Looks good!

codespan-reporting/src/term/renderer.rs
Johann150: I switched it and also the same thing in the `Bottom` variant. It's not necessarily easier now, but I added doc-comments so it should still be equally descriptive.

PR #291, opened 2020-09-20
implement clippy recommendations

Johann150:
This implements most of the recommendations Clippy gives when run for `codespan-reporting`.

There are 2 more warnings that remain:
1. `render_snippet_source` has too many arguments (max. 7 recommended, 8 present)
  &rarr;I am not sure what the `num_multi_labels` argument is for and if it can be omitted in favour of `multi_labels.len()`?
2. add a `Default` implementation for `SimpleFiles`
  &rarr;I don't think this is necessary.

### DIFF
--- a/codespan-reporting/src/term.rs
+++ b/codespan-reporting/src/term.rs
@@ -75,10 +75,8 @@ impl std::error::Error for RenderError {
 ///     pub color: ColorArg,
 /// }
 ///
-/// fn main() {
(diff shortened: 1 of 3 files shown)


brendanzab:
Yeah, I'm not sure I agree with the `Default` lint being enabled by default at all - we can add that as a global ignore to the crate. We might want to eventually add clippy to the CI if this helps.

PR #295, opened 2020-10-01
change error types

Johann150:
This unifies the `codespan-lsp` type and errors from codespan into one type that is also not as fragmented. As far as I could tell, these specific structs are not required by some lsp stuff, but please correct me if I'm wrong.

This also replaces many `Option`s in the `Files` implementation by `Result`s.

As discussed in the matrix room a little while back (cf. <https://matrix.to/#/!oaTjvHZyVsPimZCAoK:matrix.org/$Dgf8MwidQ3YMZ2hSD_PnN-aVVYhKaNY_dK5Rr3x3tCU?via=matrix.org> or view chat starting on 2020-09-17T18:39Z), this is required to accurately report errors in `term::emit`.

### DIFF
--- a/codespan-reporting/CHANGELOG.md
+++ b/codespan-reporting/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+-   All errors now use the error type `codespan_reporting::file::Error`.
+    This type also replaces the custom error type for `codespan-lsp`.
+
 ## [0.9.5] - 2020-06-24
 
 ### Changed
--- a/codespan-reporting/CHANGELOG.md
+++ b/codespan-reporting/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+There is now a [code of conduct](https://github.com/brendanzab/codespan/blob/master/CODE_OF_CONDUCT.md)
+and a [contributing guide](https://github.com/brendanzab/codespan/blob/master/CONTRIBUTING.md).
+
 ### Added
 
 -   If a label spans over multiple lines, not all lines are rendered.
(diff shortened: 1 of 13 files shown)


Johann150:
There are still some errors in the `codespan` crate. But I would rather have some opinions if this is the way to go before resolving that.

Johann150:
Because of the change in the `Files` trait, the codespan crate now relies on codespan-reporting for the error type. This causes the test of codespan without codespan-reporting to fail.
Merging this PR would make codespan depend on codespan-reporting non-optionally.

codespan/src/file.rs
```diff
@@ -2,68 +2,9 @@
...
-impl fmt::Display for SpanOutOfBoundsError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "Span out of bounds - given: {}, span: {}",
-            self.given, self.span
-        )
-    }
-}
+use codespan_reporting::files::Error;
```
brendanzab, line 7: We'll need a `#[cfg(feature = "reporting")]` on this to get rid of the error - not sure how best to do this however, as the errors are needed for the method definitions! 😬 

Also as a nitpick, I like to group the external imports and crate imports separately.

Johann150: Adding that attribute will not really help because there are then issues of the Error type not being defined (cf. <https://github.com/brendanzab/codespan/pull/295/checks?check_run_id=1205871669#step:6:14>). As I said, this would make codespan dependent on codespan-reporting *non-optionally*.

> external imports and crate imports separately

Huh, thought rustfmt would take care of that already.

Johann150: I think either we give up the independent codespan crate or this is not possible. And since there are already plans of deprecating the codespan crate anyway, the decision seems quite easy for me.

jyn514: I don't particularly care either way, I don't really maintain saltwater anymore.

Johann150: I think I'll go ahead and remove that test then.

codespan-reporting/CHANGELOG.md
```diff
@@ -7,9 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-There is now a [code of conduct](https://github.com/brendanzab/codespan/blob/master/CODE_OF_CONDUCT.md)
-and a [contributing guide](https://github.com/brendanzab/codespan/blob/master/CONTRIBUTING.md).
-
```
Johann150, line 12: This shouldn't have been moved.

PR #202, opened 2020-03-11
Tweak behaviour of line_starts to simplify Files impls

brendanzab:

PR #297, opened 2020-10-03
tweak behaviour of line_starts to simplify File impls

Johann150:
Replaces #202.

Found the bugs that were causing the tests to fail.
Easier to create another PR because I branched from the up to date master.

### DIFF
--- a/codespan-reporting/CHANGELOG.md
+++ b/codespan-reporting/CHANGELOG.md
@@ -20,6 +20,8 @@ and a [contributing guide](https://github.com/brendanzab/codespan/blob/master/CO
 
 ### Fixed
 
+-   Implementations of the `Files` trait are simplified by automatically including the
+    end of file byte index as a "line end".
 -   Empty error codes are not rendered.
 -   The locus ("location of the diagnostic") is now computed so it is always at the first
     primary label, or at the first secondary label if no primary labels are available.
--- a/codespan-reporting/examples/custom_files.rs
+++ b/codespan-reporting/examples/custom_files.rs
@@ -55,18 +55,6 @@ mod files {
         line_starts: Vec<usize>,
     }
 
-    impl File {
-        fn line_start(&self, line_index: usize) -> Option<usize> {
-            use std::cmp::Ordering;
-
-            match line_index.cmp(&self.line_starts.len()) {
-                Ordering::Less => self.line_starts.get(line_index).cloned(),
-                Ordering::Equal => Some(self.source.len()),
-                Ordering::Greater => None,
-            }
-        }
-    }
-
     /// An opaque file identifier.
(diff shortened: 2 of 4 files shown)


codespan-reporting/src/files.rs
```diff
@@ -272,14 +269,15 @@ where
 
     fn line_index(&self, (): (), byte_index: usize) -> Option<usize> {
         match self.line_starts.binary_search(&byte_index) {
-            Ok(line) => Some(line),
-            Err(next_line) => Some(next_line - 1),
+            // never return the last line index, because that is not a "real" line
+            Ok(line) => Some(std::cmp::min(line, self.line_starts.len() - 2)),
+            Err(next_line) => Some(std::cmp::min(next_line - 1, self.line_starts.len() - 2)),
```
Johann150, line 274: Here was the bug.

Johann150:
I hope this fixes what you (@brendanzab) were trying to do, but I'm not sure.

codespan-reporting/CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and a [contributing guide](https://github.com/brendanzab/codespan/blob/master/CO
 
 ### Fixed
 
+-   Implementations of the `Files` trait are simplified by automatically including the
+    end of file byte index as a "line end".
```
brendanzab, line 29: Wondering if this should be moved to 'changed'. Along with a description of how to update existing implementations of `Files`?

brendanzab: Guess I don't know if this is an improvement for end-users or not, or if it's even more confusing. Maybe @Marwes could chime in?

codespan-reporting/examples/custom_files.rs
```diff
@@ -146,8 +127,20 @@ mod files {
...
+                    .ok_or(files::Error::LineTooLarge {
+                        given: line_index,
+                        max: file.line_starts.len() - 1,
+                    })?;
+            let next_line_start = file.line_starts.get(line_index + 1).copied().ok_or(
+                files::Error::LineTooLarge {
+                    given: line_index,
+                    max: file.line_starts.len() - 1,
+                },
+            )?;
```
brendanzab, line 143: Guess I'm a bit concerned at the complexity of this? Like, how easy it is for users to make their own `Files` implementations. We might also need to update the `line_index` implementation here.

Johann150: It is bascially all the error handling that makes this so verbose. But there is no method of the Files trait that this could be moved into since `line_start` has been removed. Also `line_index` fails only if the file is missing and for other errors returns the byte index at the end of the file and that should probably not be changed.

Maybe we should also think about the opportunities this gives to implementers which goes in the direction of enabling caching and such? (*It's not a bug, it's a feature...*)

PR #312, opened 2021-01-19
rendering in US-ASCII only

### LINKED ISSUE #311: Display is broken on some platforms
Hi,

Yesterday I was playing with [Gleam](https://github.com/gleam-lang) and noticed that the errors displayed in my terminal (Cmder on Windows 10) are kinda broken.

It looks like this :
image

I guess it would be hard to have something that works on all platforms if it uses "advanced characters" (especially  on Windows). I think, it would be easier to run correctly on all platforms if diags could be displayed using only ASCII characters (ie whose range are in 0-127). I see many possible solution. One could be using an enum to instantiate a ``Char`` struct in Legacy (only ASCII) or Standard mode (make advantage of UTF-8 and Unicode) . Another one could be using cfg expression for the default value where it would change based on the platform it is compiled for. The one thing I don't know how to solve is by what should be replacing the characters that are troublesome. Especially for corners, they are single character for now and there isn't a look alike character in pure ASCII. More than one character could be used then but it would offset alignment. 

For the record, rustc seems to be only using ASCII chars.

Johann150:
Resolves #311.

### DIFF
--- a/codespan-reporting/CHANGELOG.md
+++ b/codespan-reporting/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   The `ascii-only` feature changes the defaults so that rendering uses only ASCII characters
+    (not above U+007F) for use cases that do not allow for box drawing characters, e.g.
+    terminals that do not support them.
+
 ## [0.11.0] - 2020-11-30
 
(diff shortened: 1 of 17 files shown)


Johann150:
Obsoletes #105.

brendanzab:
Tbh I'm not sure about using a feature for this. It seems like something that should be exposed by clients of the library, for example as a CLI flag or configuration option?

Johann150:
It's only one way of accessing it. Having an implementer expose it would be the nicest one.
But if I'm not completely mistaken, an end user could also add a feature flag for a dependency crate with e.g. `cargo run --features=codespan-reporting/ascii-only`, thus making this accessible if the implementers did not expose it themself.

PR #320, opened 2021-02-16
Forbid unsafe code

eduardosm:
No unsafe code seems to be needed, so we can get a green line in [cargo-geiger](https://github.com/rust-secure-code/cargo-geiger) with `#![forbid(unsafe_code)]`.


brendanzab:
Thanks!

PR #321, opened 2021-02-24
Bump version to v0.11.1

brendanzab:


Johann150 (approved):
I think it's fine to go with version `0.11.1` instead of `0.12.0`. The only breaking change is buried deep inside the library and I don't think many people actually use it. Also we are still on major version zero.

Issue #29, opened 2018-07-06
Allow for configuring the number of lines of source snippet around a label


brendanzab:
This would allow us to give users more context of problems.

Here is an example from Reason:

![](https://user-images.githubusercontent.com/1909539/29709301-ab04eac6-8940-11e7-8d2b-c65f808b6be8.png)

lpil:
This feature would be very useful for me, thank you :)

Johann150:
Is this related to #125 ?

elkowar:
Would be great if this was implemented. Is there anything specific holding this back? If not, I'd be open to trying myself at a PR for this

elkowar:
I wrote up a mock implementation in https://github.com/elkowar/codespan/commit/fa4d5f899a017a1884cbf1529ac8f85244c62183
This is most likely a pretty bad way to do this, although I'm not sure how i'd improve it - I'm really not familiar with the codebase. Any hints and thoughts would be appreciated!

Johann150:
Hmm, I think that would be the place to implement it, but not sure if its the best way, will have too look more into the code base again. But for a proper review you should open a pull request!

Johann150:
I think this issue was resolved with #331

PR #331, opened 2021-08-06
Add option for including context before and after a label

elkowar:
This PR implements #29!
I'm not sure that this is the best way to implement this, as i'm not all that deep into the codebase. If necessary, i'd appreciate any  suggestions or tips on how to improve this!

### DIFF
--- a/codespan-reporting/src/term/views.rs
+++ b/codespan-reporting/src/term/views.rs
@@ -135,6 +135,34 @@ where
                 }
             };
 
+            let lines_before = usize::min(start_line_index, self.config.before_label_lines);
+            for offset in (1..lines_before + 1).rev() {
+                let range = files.line_range(label.file_id, start_line_index - offset);
+                if let Ok(range) = range {
+                    let line = labeled_file.get_or_insert_line(
+                        start_line_index - offset,
+                        range,
+                        start_line_number - offset,
+                    );
+                    line.must_render = true;
+                } else {
+                    break;
+                }
+            }
+            for offset in 1..self.config.after_label_lines + 1 {
+                let range = files.line_range(label.file_id, end_line_index + offset);
+                if let Ok(range) = range {
+                    let line = labeled_file.get_or_insert_line(
+                        end_line_index + offset,
+                        range,
+                        end_line_number + offset,
+                    );
+                    line.must_render = true;
+                } else {
+                    break;
+                }
+            }
+
             if start_line_index == end_line_index {
                 // Single line
                 //
--- a/codespan-reporting/src/term/views.rs
+++ b/codespan-reporting/src/term/views.rs
@@ -135,28 +135,37 @@ where
                 }
             };
 
-            let lines_before = usize::min(start_line_index, self.config.before_label_lines);
-            for offset in (1..lines_before + 1).rev() {
-                let range = files.line_range(label.file_id, start_line_index - offset);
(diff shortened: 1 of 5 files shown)


codespan-reporting/src/term/views.rs
```diff
@@ -135,6 +135,34 @@ where
                 }
             };
 
+            let lines_before = usize::min(start_line_index, self.config.before_label_lines);
+            for offset in (1..lines_before + 1).rev() {
```
Johann150, line 139: Why this `min` call; shouldn't there be no context shown if `before_label_lines` is zero? The way this is implemented it would add lines 1 to the line this label starts on.

`rev`'ing the range/iterator does not change the outcome because `get_or_insert_line` will always insert the line respective of the line index. I think this could be removed.

codespan-reporting/src/term/views.rs
```diff
@@ -135,6 +135,34 @@ where
...
+                        range,
+                        start_line_number - offset,
+                    );
+                    line.must_render = true;
+                } else {
+                    break;
+                }
+            }
+            for offset in 1..lines_before + 1 {
+                let range = files.line_range(label.file_id, start_line_index + offset);
```
Johann150, line 153: I think this should start after the end of this label, i.e. from `end_line_lindex`.

Johann150 (changes requested):
Hi, thanks for your contribution. I am not sure your code does what the issue asks for. See the below more specific comments for that.

As per the Quality Standards outlined in the Contribution Guidelines I would (also) ask you to please *include some tests* for this added functionality.

codespan-reporting/src/term/views.rs
elkowar: The idea of the `min` here is to handle the edge case of being at the border of the file:
assume before_label_lines is `2`, but the label starts at the second line of the file (`start_line_index = 1`).
Now, we can only print 1 line of context before the label. The min call means that if there are less lines before the label than should be printed, we just print the amount of lines that _should_ get printed.
i.e. in the case of start_line_index = 200, the min value would be `2` - thus 2 lines of context get shown. 

This definitely needs a comment now that I think about it ^^'

Assuming that lines_before results in `0` (because the configuration sets it to 0, or we're on the first line) the loop would iterate over 1..1 - which yields no values, and thus do nothing. If I'm not missing something, this should be the desired behaviour.

`rev`ing the iterator has the purpose of making sure that if anything goes wrong with the line_range, we don't do unnecessary iterations (if line n-1 doesn't work, then line n-2 also won't work). Assuming that a failure of `line_range` is sufficiently unlikely, this could of course be removed.

codespan-reporting/src/term/views.rs
elkowar: Fixed this 👍🏻

elkowar:
I added a test to verify the implementation, and fixed the code in the view logic! 😄

codespan-reporting/src/term/views.rs
Johann150: I made some changes to this: I think this is a perfect case for using `checked_sub`. IMHO this is easier to understand than `min`.
Also in that case using `rev` actually does the opposite of what you wanted so removed that too.

Johann150 (approved):
That wraps this up. Thanks again for your contribution!

andylokandy:
When will this be released?

PR #332, opened 2021-08-08
use log10 to calculate the number of digits

Johann150:
This way it is not necessary to have a loop and should be a little bit more efficient.


brendanzab:
Oooh, super cool! 🎉

brendanzab:
Seems like we are waiting on https://github.com/rust-lang/rust/issues/70887 to avoid the float conversion

Johann150:
Ah good catch, added a comment on that.

PR #338, opened 2021-10-03
Fix typo in ci.yml

Sematre:
The name of those ci steps suggest that they run ``cargo test`` when they actually run ``cargo check``.


Johann150:
Thanks!

PR #341, opened 2021-11-16
Added generic iterator methods

yankana:
It's unnecessary to require allocated vectors instead of just any iterator. With this change, you can use `with_labels_iter` and `with_notes_iter` in the example code (and in other code) with arrays instead of vectors to avoid extra heap allocations. The other methods were left untouched so this isn't a breaking change.

### DIFF
--- a/codespan-reporting/src/diagnostic.rs
+++ b/codespan-reporting/src/diagnostic.rs
@@ -182,10 +182,22 @@ impl<FileId> Diagnostic<FileId> {
         self.labels.append(&mut labels);
         self
     }
+    
+    /// Add some labels to the diagnostic.
+    pub fn with_labels_iter(mut self, labels: impl IntoIterator<Item = Label<FileId>>) -> Diagnostic<FileId> {
+        self.labels.extend(labels);
+        self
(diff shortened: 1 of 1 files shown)


Johann150:
Thanks for your contribution!

Since `Vec<String>` is compatible with  `impl IntoIterator<Item = String>`, and likewise for the other function, I think we could even just replace the existing methods instead of creating new ones.

yankana:
> Since `Vec<String>` is compatible with `impl IntoIterator<Item = String>`, and likewise for the other function, I think we could even just replace the existing methods instead of creating new ones.

This would be a breaking change because changing a method from using a concrete type to a generic/opaque one would break code that relies on the compiler inferring types to be vectors. Example:
```rust
// `collect::<Vec<String>>` is inferred here, but with the new update, the compiler complains
let diagnostic = Diagnostic::error().with_notes(string_iter.collect());
```

kaleidawave:
Added a commit that fixed a syntax error. Checked locally and all works okay. New methods will not create a break change.

Thanks for the change!

Issue #293, opened 2020-09-26
Add the ability to make multiline suggestions/notes


RDambrosio016:
Hello!

I have been working on a large project and a lot of the errors have suggestions attached to them, however, i would greatly appreciate being able to make suggestions like rustc_errors can do. Example:
```
error: unknown start of token: \u{37e}
 --> main.rs:2:30
  |
2 |    println!("Hello, world!");
  |                             ^
help: Unicode character ';' (Greek Question Mark) looks like ';' (Semicolon), but it is not
  |
2 |    println!("Hello, world!");
  |                             ^
```
Being able to make suggestions which point to some lines in the source code but change it and label the changes would be a great feature, however, thinking of a good API to provide this will not be an easy task. Another thing i have considered is potentially exposing the renderer API to allow users to tack on their own render logic, which would help me immensely.

brendanzab:
Yeah, this has been requested before (See #285)! Perhaps this is something we should have a look at.

For this specific instance, I'm wondering if using the notes section would help. Eg.

```
error: unknown start of token: \u{37e}
  ┌─ main.rs:2:30
  |
2 |    println!("Hello, world!");
  |                             ^
  = Unicode character ';' (Greek Question Mark) looks like ';' (Semicolon), but it is not
```

But I realise there are other instances where this is not as helpful, say with lifetime diagnostics, where you have to step through a sequence of related locations.

It would also be good to think about how this maps onto the language server protocol, but we might not want to limit ourselves by it. Eg. I tend to find lifetime errors easier to understand in the terminal, rather than using rust-analyzer.

RDambrosio016:
As with most pretty error stuff, the logic will be lost in the LSP, which is not necessarily bad, i agree with you in saying cli is way better for some errors, i dont think LSP should limit codespan when it comes to features. The example was the first i could think of, it is obviously very simplistic.

Johann150:
I have not looked at the commits that were made, but maybe #176 could be interesting in this context?

RDambrosio016:
Allowing users to write their own render logic and plug into codespan's renderer would be infinitely helpful when it comes to more specific diagnostics

Aapuji:
Hello, I was wondering what the status is for this issue? Thanks

kaleidawave:
I think some of this has been implemented and discussed in the spade fork: https://github.com/brendanzab/codespan/issues/372 other that there is no open PR implementing this.

[...]

PR #346, opened 2022-07-27
Allow adding labels to notes

nolanderc:
Attempts to address #293 by extending each note with an optional list of labels:
```
error: cycle detected when evaluating constant `A`
  ┌─ one_line.rs:1:1
  │
1 │ const A: u32 = B;
  │ ^^^^^^^^^^^^^^^^^
  │
  = ...which requires evaluating constant `B`...
  ┌─ one_line.rs:2:1
  │
2 │ const B: u32 = A;
  │ -----------------
  │
  = ...which requires evaluating constant `A`
  ┌─ one_line.rs:1:1
  │
1 │ const A: u32 = B;
  │ -----------------
```

### DIFF
--- a/codespan-reporting/examples/custom_files.rs
+++ b/codespan-reporting/examples/custom_files.rs
@@ -9,7 +9,7 @@
 //! cargo run --example custom_files
 //! ```
 
-use codespan_reporting::diagnostic::{Diagnostic, Label};
+use codespan_reporting::diagnostic::{Diagnostic, Label, Note};
 use codespan_reporting::term;
 use codespan_reporting::term::termcolor::{ColorChoice, StandardStream};
 use std::ops::Range;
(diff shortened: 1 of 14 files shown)


brendanzab:
Cool, thanks so much! Apologies for the lackluster response time. If labels are attached to the notes, I wonder if it would make sense to indent them to demonstrate the hierarchy?

```
error: cycle detected when evaluating constant `A`
  ┌─ one_line.rs:1:1
  │
1 │ const A: u32 = B;
  │ ^^^^^^^^^^^^^^^^^
  │
  = ...which requires evaluating constant `B`...
      ┌─ one_line.rs:2:1
      │
    2 │ const B: u32 = A;
      │ -----------------
      │
      = ...which requires evaluating constant `A`
      ┌─ one_line.rs:1:1
      │
    1 │ const A: u32 = B;
      │ -----------------
```

That might look a bit weird though…

If not I wonder if another solution could be to allow for a series of trailing labels attached to the top level diagnostic that are always rendered in order. 🤔

No worries if you are on other things at the moment.

kaleidawave:
I *think* this implements #293

If there is still interest I can investigate getting this in the next release

PR #350, opened 2023-05-21
update dependencies in codespan-reporting

LighghtEeloo:
Ran tests, and everything seems to work.


abs0luty:
The repo seems to be abandoned, unfortunately :(.

Johann150:
The minimum supported rust version is currently 1.46.0 which does not support `edition = 2021` which would be required by rustyline 11.

PR #355, opened 2023-10-17
include LICENSE file in published crates

decathorpe:
The terms of the Apache-2.0 license include a requirement that (re)distributed sources contain a copy of the license text. This is a problem for Linux distributions which need to package the crates from this project (and technically, it's also a problem for redistributing sources via crates.io).

This PR includes a symlink to the LICENSE file in all three published crates, which should make `cargo publish` inlude this file when publishing to crates.io. If this project is published from a Windows based machine, full copies of the file might be needed instead.


danakj:
I filed https://github.com/brendanzab/codespan/issues/365 which this PR will resolve.

anforowicz:
@Johann150, can you please take a look?  This PR will help resolve Chromium bug at https://crbug.com/378272935

kaleidawave:
As explained in #365, other repositories need a LICENSE file for reference (as they don't read the manifest file). I see this PR adds symlinks for the sub-crates to reference the main license (thus staying in sync), which should help users of this crate.

I believe the release is done on a Linux machine but I think using symlink should also be fine on a Windows machine (I have had success using symlinks with Git on my windows machine).

Thanks for this improvement!

decathorpe:
Thanks!

PR #369, opened 2024-11-28
unicode-width: allow 0.2 to be selected

nagisa:
Although the lockfile remains locked to 0.1.14 in this project as there are many other dependencies (e.g. rustyline) which depends on ^0.1, but for dependents on `codespan-reporting` specifically it allows to update to 0.2 without duplicating dependencies.


kaleidawave:
Change in semver still allows `0.1` and `unicode-width` is at `0.2` so this should be fine to merge.

PR #373, opened 2025-02-08
Add strict clippy lints and any changes after "auto-fix"

kaleidawave:
- Add strict clippy lints
- Add any changes after "auto-fix"
- Add crates-release-action for releasing


kaleidawave:
Getting CI test suite errors but as they are only about dependencies and for an old version and nightly version I think it is fine

PR #387, opened 2025-06-18
Make rendering color agnostic

### LINKED ISSUE #386: Consider making rendering color agnostic
Instead of taking in a WriteColor, the renderer should take in WriteStyle. this decouples the rendering logic from any specific color theme, making it trivial to support new themes (dark, light, high-contrast) or entirely different output formats (HTML, terminal, PDF) simply by plugging in a different StyleWriter implementation.

urisinger:
Resolves #386

This adds a new `WriteStyle` trait for styling that replaces WriteColor. 

Right now, the only public APIs I’ve changed are the ones for `Config` and `emit`: I removed `Style` from `Config` (because a renderer doesn’t always need one) and added a `style` parameter to `emit`.

Using the API is very simple, just implement the `WriteStyle` trait, and then pass your writer to the renderer. I’ve implemented a minimal HTML backend [here](https://github.com/urisinger/tevel-dashboard/blob/main/backend/compiler/src/diagnostics.rs), based on [readme preview](https://github.com/brendanzab/codespan/blob/master/codespan-reporting/examples/readme_preview.rs).


urisinger:
> Oooh I have run into that issue before... I don't think it was on this repository and I cannot remember how I fixed it. I can only think: double check you are on the latest and run `cargo build` etc with `--locked`

works now, even managed to built the example svg: example svg

urisinger:
tests are now passing. should this be locked behind the termcolor feature just like before? I think i might be able to get this to work without it.

urisinger:
the pr is ready to review. im not quite sure about the api yet, maybe we can add some helper functions

codespan-reporting/src/term/config.rs
```diff
@@ -215,6 +235,63 @@ impl Default for Styles {
...
+#[cfg(feature = "termcolor")]
+impl<'a, W> StylesWriter<'a, W> {
+    pub fn new(writer: W, style: &'a Styles) -> Self {
+        Self { writer, style }
+    }
+}
+
+#[cfg(feature = "termcolor")]
+#[cfg(feature = "std")]
+impl<'a, W: WriteColor> io::Write for StylesWriter<'a, W> {
```
kaleidawave, line 253: Nice! So this `impl` ensures all existing uses will not break and this update will be backwards compatible?

kaleidawave:
Looks good, have a few queries but after they are resolved and I have tested this on one of my own projects I will merge.

codespan-reporting/src/term/config.rs
urisinger: nope, you will have to create a StylesWriter instead of a ColorWriter, but its just a two line change with
```rust
let styles = Styles::default();
let styles_writer = StylesWriter::new(writer, &styles);
```

codespan-reporting/src/term/config.rs
```diff
@@ -104,51 +108,51 @@ pub enum DisplayStyle {
 pub struct Styles {
```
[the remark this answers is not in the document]
urisinger: i used it to refactor and forgot to make it public, they definitely should be public

urisinger:
@kaleidawave any progress?

urisinger:
> Hi @urisinger , apologies have been busy recently.
> 
> As I have only recently become a maintainer and still don't have a lot of knowledge of the codebase and this change is breaking I have put this _on the back-burner_.
> 
> I have found a few changes that would be needed in external repositories, I will add some more as I test others
> 
> ```
> gfx-rs/wgpu naga/src/error.rs:50:17 & naga/src/span.rs:330:20
> bevyengine/naga_oil src/compose/error.rs:295:20
> ```
> 
> But other repositories: `dtolnay/cxx` and `EmbarkStudios/toml-span` already work 🎉
> 
> I can merge this soon, but need 3 things:
> 
>     1. I have a few things to add to my change to CI PR [Update and improve CI and testing #388](https://github.com/brendanzab/codespan/pull/388) but it should be ready soon? If this PR passes under that I will be more confident about merging it.
> 
>     2. To investigate whether this PR can fix [`term::emit`'s `writer` argument does not behave additively wrt features. #389](https://github.com/brendanzab/codespan/issues/389) (or be modified not to run into the current issue)?
> 
>     3. For a short piece of text on what users of the library **need to modify** when updating. Something I can put into the changelog/release notes for `0.13.0` update
> 
> 
> Thanks!

[this commit](https://github.com/brendanzab/codespan/pull/387/commits/fbe2b181defc88db9b808baa33ad126a8811db05) fixed [#389](https://github.com/brendanzab/codespan/issues/389), it provided a blaket implementation of StylesWriter for ColorWriter that uses the default Color config. It might be possible to implement StylesWriter for a regular Writer if the term feature is not turned on.

kaleidawave:
Hi! Apologies I am only just getting round to this. I have merged a PR since. I updated the branch and made a few minor updates because I removed the `lazy_static` dependency. 

I am now running the tests and seem to get some issues on this PR. I do not totally understand them, so maybe you can help?

https://github.com/brendanzab/codespan/actions/runs/18048940552/job/51366544202?pr=387

If it is completely my fault, let me know and I will take a closer look if I made an error.

Once CI is green, I will merge. Definitely this time 😅.

urisinger:
I think i accidently made a change to the color scheme, not sure what exactly.

urisinger:
are there any tests that run with termcolor feature disabled? @kaleidawave

kaleidawave:
Yep, commit it here (I think that works). I can copy that diff over to another branch

urisinger:
also the update is in no way backwards compatible, it needs to be a new version

PR #403, opened 2025-10-15
Regenerate Cargo.lock with 0.13.0 versions

dtolnay:
6a28ecfc6b44a84393a1e2c7240432855eaaa094 updated 3 Cargo.toml files, but did not check in the resulting change to Cargo.lock.


kaleidawave:
Thanks. Would have be included normally but the commit part of my workflow broke and forgot to do it when adding the manual correction.

PR #412, opened 2026-03-14
Add html writer

urisinger:
This new html renderer makes it easy to render errors as html documents and serves as an example for similar renderers. 
Makes readme_preview use it and has a readme example too.


kaleidawave:
Cool! I can see some uses cases for this. How does this compare to the SVG writer?

Will add it before the next release.

urisinger:
> Cool! I can see some uses cases for this. How does this compare to the SVG writer?
> 
> Will add it before the next release.

Its almost the same as the svg writer, just as a public interface.

Issue #414, opened 2026-08-17
Hyperlnk for snippet_locus function


saeedark:
Hi,
Thanks for this lib.

I would be practical to add hyperlink for filenames. It lets users to navigate to the error location faster.

thanks

PR #415, opened 2026-08-18
feat: reporting with hyperlinks

saeedark:
Adding hyperlink for reporting can help to easier navigate source codes.

#414


kaleidawave:
Thanks for this change!

I assume that this is for 'integrated terminals' in IDEs?

However, I believe VS Code and Zed already automatically add links for path like output. ([see VS Code feature](https://code.visualstudio.com/docs/terminal/basics#_links)). Do you have example of a program where the link is not automatically detected and thus would be benefit from this feature?

> I was thinking about this recently, it would be nice if you could have an option in a terminal where clicking on a link like this could be configured to spawn `hx $dest` in a split window?

saeedark:
> I assume that this is for 'integrated terminals' in IDEs?

No, actually It's for normal terminals. 

> However, I believe VS Code and Zed already automatically add links for path like output. ([see VS Code feature](https://code.visualstudio.com/docs/terminal/basics#_links)). Do you have example of a program where the link is not automatically detected and thus would be benefit from this feature?

No, I was typesetting a document file with Typst in a terminal and I really wanted this feature!

> I was thinking about this recently, it would be nice if you could have an option in a terminal where clicking on a link like this could be configured to spawn hx $dest in a split window?

That's my work flow. However Implementing this won't be easy. See [this](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname) and [this](https://github.com/rust-lang/annotate-snippets-rs/issues/366).  I thought about adding a URI getter function to Files trait.  This new commit will do that. basically users should implement it themselves.

